PR #36, opened 2017-07-25
Consolebot

aaronschachter:
* Restores the Consolebot script, updates README
* Begins test coverage a la https://github.com/DoSomething/gambit/pull/933
* Returns the `req.outboundMessage` as a reply property in our `/receive-message` response

@rapala61 and I have been chatting whether to make a single POST `/messages` endpoint that would accept inbound messages (which need a reply), our outbound send messages (Front support messages), and our outbound imported messages (copies of c.io broadcasts). Returning the `reply` Message we create for an inbound message feels like a nice touch, and also makes me still consider separate endpoints...  although it probably makes sense to include the initial `inbound` Message model we create from the incoming parameters...


lib/consolebot.js
```diff
@@ -0,0 +1,61 @@
...
+    });
+    this.readline.setPrompt(`${this.options.prompt} `.bold);
+    this.readline.on('line', input => this.post(input));
+    this.readline.on('close', () => process.exit(0));
+  }
+  post(inboundMessageText) {
+    return superagent
+      .post(this.options.url)
+      // TODO: Add authentication to POST /receive-message
+      // .set('x-gambit-api-key', this.options.apiKey)
```
aaronschachter: Refs #38

PR #40, opened 2017-07-27
Simple implementation to receive and reply to Mobile Commons messages 

aaronschachter:
I've created a new mData that posts to `/receive-message?medium=mobilecommons` and sends the Conversations API response back to the User via Mobile Commons.

This branch adds a check for a `medium` query parameter. When set to `mobilecommons`, set our `req.userId` to the `req.body.phone`, and set our `req.inboundMessageText` as the `req.body.keyword` if it exists.

An mData can be configured to render the response from its Web Service URL with Liquid -- the Gambit Conversations mData sends the [`reply.text` response of its POST `/receive-message` request](https://github.com/DoSomething/gambit-conversations/tree/master/documentation#response).

screen shot 2017-07-26 at 5 10 09 pm

We can also customize the parameters the mData sends. What's really nice about this approach is we only get the parameters we define, not the entire mData payload:

screen shot 2017-07-26 at 5 10 22 pm

A kicker here is we're cutting Blink out of the picture, since the mData is waiting on the request to the Conversations API to send the reply back. Looking to investigate if it's possible to use our Mobile Commons Twilio API creds for sending the reply back, like how we do for our current Gambit Campaigns `chatbot` endpoint, vs relying on the mData waiting for a response to send.

### DIFF
--- a/lib/middleware/receive-message/message-inbound.js
+++ b/lib/middleware/receive-message/message-inbound.js
@@ -4,6 +4,9 @@ const helpers = require('../../helpers');
 
 module.exports = function createInboundMessage() {
   return (req, res, next) => {
+    if (req.inboundMessageText) {
+      req.userCommand = req.inboundMessageText.trim();
+    }
     req.conversation.createInboundMessage(req.inboundMessageText)
       .then((message) => {
         req.inboundMessage = message;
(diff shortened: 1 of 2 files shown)


aaronschachter:
yep, I did! Thanks!

PR #41, opened 2017-07-27
Auth

### LINKED ISSUE #38: Authentication
Doesn't exist yet

rapala61:
This PR Adds Basic authentication to the `send-message` and `receive-message` routes.

## How to test
- 👀 
- Run `npm test`
- Use the consolebot to test conversation.

## Relevant tickets
- Fixes #38


.env.example
```diff
@@ -2,8 +2,6 @@ DS_GAMBIT_CAMPAIGNS_API_KEY=totallysecret
 DS_GAMBIT_CAMPAIGNS_API_BASEURI=http://ds-mdata-responder-staging.herokuapp.com/v1
 DS_GAMBIT_CAMPAIGNS_SYNC=anything
 
-SLACK_BOT_TOKEN=totallysecret
```
aaronschachter: Mistake?

.env.example
```diff
@@ -14,3 +12,6 @@ FB_PAGE_ACCESS_TOKEN=totallysecret
 FRONT_API_TOKEN=totallysecret
 FRONT_API_SECRET=totallysecret
 FRONT_API_SUPPORT_CHANNEL=cha_290x
+
+BASIC_AUTH_NAME=puppet
```
aaronschachter: What do you think about prefixing this with something like `DS_GAMBIT_CAMPAIGNS_API_` to spell it out that these vars are used to authenticate our Gambit Campaigns requests?

.env.example
rapala61: I deleted because I didn't find any occurrence of this variable in the project.

.env.example
rapala61: Mmm, actually we are authenticating incoming requests to the Gambit Conversations API.

.env.example
aaronschachter: Ah, you're right. I've been bad at keeping this file up to date when I was adding the send message, it's defined as `SLACK_API_TOKEN` [here](https://github.com/DoSomething/gambit-conversations/blob/c003af0be487a4941ecb9b860ebfe7752e68f285/lib/slack.js#L7). Mind updating this line with the correct name?

.env.example
aaronschachter: oh man. these names. yes, you're totes right -- `DS_GAMBIT_CONVERSATIONS_API_`

PR #45, opened 2017-08-08
Added newrelic to Conversations API

rapala61:
## How to review:
- 👀


server.js
```diff
@@ -31,7 +37,7 @@ restify.serve(app, MessageModel);
 const db = mongoose.connection;
 db.on('error', () => {
   // TODO console.log has to be replaced by other development logging library: Winston?
-  // console.error.bind(console, 'connection error:')
+  // console.error.bind(console, 'connection error:');
```
aaronschachter: The [`heroku-logger`](https://github.com/DoSomething/gambit-conversations/blob/d1b2f9abc6c6f673f463743dce08a82cf0ac1a5e/package.json#L38) has been added to this repo -- although it's hard to say it's anymore more useful that Winston without integrating with Logstash.

rapala61: It makes sense to use `heroku-logger`. From what I can see, the message part of the log will have the payload we pass as the first argument. I liked Winston because it was asynchronous, eslint compliant, and great for free form troubleshooting.

Issue #31, opened 2017-07-24
Save attachments to Message model


aaronschachter:
We'll need to store any attachments that exist for inbound and outbound messages.

For outbound messages, we currently don't have any fields in Contentful to use for attachments on an outbound template. However, an agent could include attachments via Front that we'd want to save to our `Message` model.

rapala61:
I can take on this. How does that attachment URL gets sent to the user? As part of the text response?

I need more info to get to work on this as I'm not familiar with how Front includes this URL. This looks like an specific parsing middleware/function for incoming Front messages, but it seems related to the incoming messages with mms from Twilio too. Should this functionality be included in the middleware we are thinking about implementing to deprecate `Adapters`? Ultimately, the `Message`  will be updated with an `attachment` property, but how to get there needs a little more fleshing out.

aaronschachter:
When thinking through outbound attachments, this issue turns into a really big one which would also mean a really big PR if it was in one branch (e.g. parsing Front)

@rapala61 Maybe let's just start with storing inbound messages attachments from Twilio messages? Which means this is blocked by https://github.com/DoSomething/gambit-conversations/pull/50/files#r132314865

rapala61:
This task addresses just the attachment problem, but I know we want to port the whole of the functionality from Adapters into Convo API, should this task (#31) be included in the bigger [Adapters -> Convo API migration](https://github.com/DoSomething/gambit-conversations/issues/53)? The many similar issues are a little confusing to scope separately @aaronschachter

aaronschachter:
this issue was created to document that we need both inbound + outbound messages.

I think it makes sense to add storing inbound messages along with #31 but outbound should be a separate PR to making reviewing easier. I'm open to renaming this issue as inbound message storage and opening an outbound message storage @rapala61 if that would be easier...

rapala61:
@aaronschachter I think renaming and creating a separate for outbound messages is a good idea.

rapala61:
@aaronschachter 
This is an example of an incoming payload we should get from Blink (redacted for privacy):
```
{
  "ToCountry": "US",
  "MediaContentType0": "image/png",
  "ToState": "",
  "SmsMessageSid": "MM09a8f657567f807443191c1e7exxxxxx",
  "NumMedia": "1",
  "ToCity": "",
  "FromZip": "10010",
  "SmsSid": "MM09a8f657567f807443191c1e7exxxxxx",
  "FromState": "NY",
[...]

PR #50, opened 2017-08-09
Support passing a From number to Twilio Messages API

aaronschachter:
* Adds optional `TWILIO_FROM_NUMBER` config variable to post a message using a [From number instead of a Messaging Service Sid](https://www.twilio.com/docs/api/rest/sending-messages#post-parameters-conditional). This allows us to post messages using our Mobile Commons Twilio API credentials by passing our shortcode or test longcode as the `From`.
    * Adds `/config/lib/twilio` to define Twilio config vars 

* Removes check for a `mobilecommons` query parameter, as we don't need an mData (#40) to post Mobile Commons messages to Conversations API per https://github.com/DoSomething/blink/pull/97

* Creates Conversation with medium `sms` instead of `twilio` for SMS Conversations

* `/receive-message` sets our Conversation medium as SMS by checking for expected Twilio parameters, so we're ready for Blink to sending messages received by our Mobile Commons test number: https://github.com/DoSomething/blink/issues/102
    * This essentially deprecates our Twilio route in Gambit Adapters, as Blink is posting directly to Conversations (not Adapters). I think this makes sense to do, since Conversations API will return an error if for whatever reason it can't process the message it receives.

### DIFF
--- a/config/lib/twilio.js
+++ b/config/lib/twilio.js
@@ -1,0 +1,8 @@
+'use strict';
+
+module.exports = {
+  accountSid: process.env.TWILIO_ACCOUNT_SID,
+  authToken: process.env.TWILIO_AUTH_TOKEN,
+  fromNumber: process.env.TWILIO_FROM_NUMBER,
+  messagingServiceSid: process.env.TWILIO_MESSAGING_SERVICE_SID,
+};
--- a/lib/middleware/receive-message/params.js
+++ b/lib/middleware/receive-message/params.js
@@ -9,18 +9,6 @@ module.exports = function params() {
     const body = req.body;
     req.inboundMessageText = body.text;
 
-    if (req.query.medium === 'mobilecommons') {
-      req.platform = 'mobilecommons';
-      req.userId = body.phone;
-
-      const keyword = req.body.keyword;
-      if (keyword) {
-        req.inboundMessageText = keyword;
-      }
-
-      return next();
-    }
-
     if (body.slackId) {
       req.platform = 'slack';
       req.userId = body.slackId;
@@ -29,9 +17,10 @@ module.exports = function params() {
       return next();
     }
 
-    if (body.phone) {
-      req.platform = 'twilio';
-      req.userId = body.phone;
+    if (body.MessageSid) {
+      req.platform = 'sms';
+      req.userId = req.body.From;
+      req.inboundMessageText = req.body.Body;
 
       return next();
     }
--- a/lib/middleware/receive-message/params.js
+++ b/lib/middleware/receive-message/params.js
@@ -19,8 +19,8 @@ module.exports = function params() {
 
     if (body.MessageSid) {
       req.platform = 'sms';
-      req.userId = req.body.From;
-      req.inboundMessageText = req.body.Body;
+      req.userId = body.From;
+      req.inboundMessageText = body.Body;
 
       return next();
     }
--- a/lib/middleware/receive-message/params.js
+++ b/lib/middleware/receive-message/params.js
@@ -21,6 +21,8 @@ module.exports = function params() {
       req.platform = 'sms';
       req.userId = body.From;
       req.inboundMessageText = body.Body;
+      // TODO: We need to save attachments here if we're ditching Gambit Adapters and Blink is
+      // posting directly to this /receive-message route.
 
       return next();
     }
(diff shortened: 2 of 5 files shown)


lib/middleware/receive-message/params.js
```diff
@@ -29,9 +17,12 @@ module.exports = function params() {
...
 
-    if (body.phone) {
-      req.platform = 'twilio';
-      req.userId = body.phone;
+    if (body.MessageSid) {
+      req.platform = 'sms';
+      req.userId = body.From;
+      req.inboundMessageText = body.Body;
+      // TODO: We need to save attachments here if we're ditching Gambit Adapters and Blink is
+      // posting directly to this /receive-message route.
```
aaronschachter: Seems it's time to decide to take or leave Gambit Adapters. 

If we want to keep it, Blink will need to POST to Adapters, which posts to Conversations, which posts to our medium (Twilio, Slack, etc). If there's an error in the Conversations API request, it needs to tell Adapters which tells Blink so it can retry. This feels like a lot of work for the sake of our Conversations API not handling how to parse an inbound message.

At the same time -- Twilio sends us a URL that redirects to an image, so we'll have to [port this code in Adapters over too](https://github.com/DoSomething/gambit-adapters/blob/562e8bc1085b4ee960ef86f9005802c713b791e9/app/routes/twilio.js#L25) into Receive Message middleware, to inspect and save attachments on inbound messages. 

I'm inclined to deprecate Adapters for the sake of less moving pieces. What do you think @rapala61 @sergii-tkachenko ?

config/lib/twilio.js
```diff
@@ -0,0 +1,8 @@
+'use strict';
+
+module.exports = {
+  accountSid: process.env.TWILIO_ACCOUNT_SID,
+  authToken: process.env.TWILIO_AUTH_TOKEN,
+  fromNumber: process.env.TWILIO_FROM_NUMBER,
```
aaronschachter: We'll need to set this to our test longcode number on Conversations Staging to send replies to the received messages that Blink will be posting soon.

lib/middleware/receive-message/params.js
aaronschachter: Refs #31

rapala61: I like the idea of creating a robust middleware that can massage the incoming POST requests to `receive-message` based on the medium/source. We could create another blink endpoint that 3rd party apps can POST to which then can be routed to the `receive-message`. We can add a query param to the webhook URL at the 3rd party app level to aid in parsing it `?medium=slack` for now. As we scale, and we start experiencing a lot more traffic, we can open a queue per service, but the consumers would POST to `receive-message`.

config/lib/twilio.js
rapala61: Is this one of those ENV variables that should explode the app if its not set? If so I think we should check their existence.

lib/middleware/receive-message/params.js
aaronschachter: @rapala61 talked through this in Slack -- yes, let's have Blink post directly to our `/receive-message` endpoint, deprecating the use of Gambit Adapters for receiving Twilio messages.

We'll want to port [the code over from Adapters to save attachments in an inbound Twilio message.](https://github.com/DoSomething/gambit-adapters/blob/562e8bc1085b4ee960ef86f9005802c713b791e9/app/routes/twilio.js#L25)

config/lib/twilio.js
aaronschachter: It's questionable whether to explode the app... but probably makes sense. In order to deliver a reply message to a SMS Conversation, we need either the `TWILIO_FROM_NUMBER` or `TWILIO_MESSAGING_SERVICE_SID` config vars set.

An edge-case to make the argument for not exploding the app is if someone wanted to add Rivescript and test locally with consolebot -- we wouldn't be making any POST requests to Twilio in this case.

rapala61: I'm not super familiar with Twilio message properties yet. Why do we need `TWILIO_FROM_NUMBER ` **OR** `TWILIO_MESSAGING_SERVICE_SID ` to reply back to an incoming message? I thought we always needed a From number since the SID was just an identifier of that particular message.

aaronschachter: https://www.twilio.com/docs/api/rest/sending-messages#post-parameters-conditional

Our Twilio test code has been working without passing the From number thus far by passing a Messaging Service SID (created/maintainable through our Twilio console)

rapala61: Ah, thanks. I confused the Account SID with the Message SID.

PR #91, opened 2017-08-28
Rename DB properties

aaronschachter:
Refs #39:
* Rename Conversation `medium` to `platform`
* Rename Conversation `userId` to `platformUserId`
* Rename Messages `conversation` to `conversationId`

PR #92, opened 2017-08-29
Receive Message middleware cleanup

### LINKED ISSUE #86: Return error from Gambit Campaigns /receive-message
When posting Signup Messages to the Gambit Campaigns service, we need to return any errors we get back, like 504's, for Blink to retry failed requests.

See [Send Gambit Campaigns Reply middleware](https://github.com/DoSomething/gambit-conversations/blob/0.3.1/lib/middleware/receive-message/send-gambit-campaigns-reply.js#L37).

aaronschachter:
Refactors our [POST `/receive-message`](https://github.com/DoSomething/gambit-conversations/blob/master/documentation/endpoints/receive-message.md) middleware to use new helper functions that create a reply `Message` to our inbound `Message`, post the message text to the Conversation platform, and return the HTTP response. This is similar to how the Gambit Campaigns [`lib/conversations`](https://github.com/DoSomething/gambit/blob/4.6.3/lib/conversation/replies.js) code works (introduced in https://github.com/DoSomething/gambit/pull/926).

Before:
* When making a Rivescript reply request, we'd set a `req.reply` object, with its `brain` property set to the coded Rivescript reply (stored in the `/brain`).
* Each middleware function would check the `req.reply.brain` to determine if it should set the `req.reply.template` -- the message template we want to use to reply.
* After executing all of the middleware functions:
   * Render the `req.reply.template` text (sometimes using `req.campaign` variables, depending on the template)
    * Create the outbound reply `Message`
    * Send the successful HTTP 200 POST response

Now:
If a middleware function is responsible for setting the template, call one of the helper functions to send back the reply:
* `helpers.sendReply(req, res, messageText, messageTemplate)`
* `helpers.sendReplyWithCampaignTemplate(req, res, messageTemplate)` -- calls `sendReply` with text `req.campaign.templates[messageTemplate]` 
* `helpers.sendReplyForCampaignSignupMessage(req, res)` -- posts `req.inboundMessage` to the Gambit Campaigns `/receive-message` endpoint, and calls `sendReply` with POST response text
    * Fixes #86 -- returns a Gambit Campaigns error for errors

Details:
* Removed `conversation-set-campaign.js` -- Instead of a separate middleware, set the Conversation Campaign from the few that should set it:
    * `campaign-menu` -- Set the Conversation Campaign to the random Campaign we're asking for Signup
   * `campaign-keyword` -- TODO: Make sure this is happening
* Removed `message-outbound.js` -- Outbound reply `Message` creation is handled in `helpers.sendReply`
* Combined `rivescript-reply` and `send-rivescript-reply` into single `rivescript` middleware
* Removed `send-campaign-message.js` -- deprecated by middleware calling `helpers.sendReplyWithCampaignTemplate`.
   * We set the default `noCampaignMessage` to send when a Conversation doesn't have a `campaignId` set in `campaign-current`
* Refactored `/lib/rivescript` to simplify (refs https://github.com/DoSomething/gambit-conversations/pull/91#discussion_r135578448), which was the start of this massive "wait, is this working this way" refactor. Seemed like the appropriate time to take it on as we're still far from going live in production.

Testing:
* I'll deploy this branch to staging so we can Slack with `@gambit-staging` to verify conversation flow.


app/models/Conversation.js
```diff
@@ -48,16 +48,18 @@ conversationSchema.statics.createFromReq = function (req) {
...
+conversationSchema.statics.getFromReq = function (req) {
+  const query = {
+    platformUserId: req.platformUserId,
+    platform: req.platform,
+  };
+  logger.trace('Conversation.getFromReq', query);
 
   return this.findOne(query)
-    .then(convo => convo)
+    .then(conversation => conversation)
```
deadlybutter: What's the point of this line?

app/routes/receive-message.js
```diff
@@ -36,38 +30,31 @@ router.use(createConversationMiddleware());
...
-// Includes catchall for when no Campaign has been set, and no Rivescript reply matched.
-router.use(sendCampaignMessageMiddleware());
+// If our last outbound template was not for the Campaign, prompt to continue Campaign Completion.
+router.use(continueCampaignMiddleware());
 
-// Create our outbound message and send the reply back to sender (if conversation isn't paused).
-router.use(outboundMessageMiddleware());
-router.post('/', (req, res) => helpers.sendResponse(req, res));
+// If we haven't matched anything yet, send this message to Gambit Campaigns for a reply.
+router.post('/', (req, res) => helpers.sendReplyForCampaignSignupMessage(req, res));
```
deadlybutter: Hm, the function name and the comment don't entirely match here. It sounds like the function is specific to sign ups, but the comment is just saying "reply". Is this just a lingo thing I'm not aware of?

lib/helpers.js
```diff
@@ -73,6 +74,90 @@ module.exports.sendResponseWithStatusCode = function (res, code = 200, message =
...
+/**
+ * Creates and sends outbound-reply Message with given messageText and messageTemplate.
+ * @param {object} req
+ * @param {object} res
+ * @param {string} messageText
+ * @param {string} messageTemplate
+ */
+module.exports.sendReply = function (req, res, messageText, messageTemplate) {
+  logger.debug('sendReply', { messageText, messageTemplate });
+  let outboundMessage;
```
deadlybutter: Nitpicky, should this have a default value like null?

lib/helpers.js
```diff
@@ -73,6 +74,90 @@ module.exports.sendResponseWithStatusCode = function (res, code = 200, message =
...
+  return req.conversation.createOutboundReplyMessage(messageText, messageTemplate)
+    .then((message) => {
+      outboundMessage = message;
+      logger.debug('createOutboundReplyMessage', { messageId: message._id.toString() });
+
+      if (!outboundMessage.text) {
+        return true;
+      }
+
+      return req.conversation.postMessageToPlatform(outboundMessage);
```
deadlybutter: Following this promise, in general, is a bit weird. The return statements in this first function don't propagate downwards so not sure why they're being used. Additionally, I'm worried about the scoping being used. It doesn't feel appropriate for a promise function to modify a value in a higher scope. I think the proper way to approach this would be having the function return a promise. In the first function, figure out the outbound message, return that. In the second function, accept that message as a parameter and send that to the proper place.

Finally, when reading this it reads as if the message is posted twice, which seems weird? I'm judging this based on `req.conversation.postMessageToPlatform(outboundMessage);` and `return res.send({ messages });` both being called.

lib/helpers.js
```diff
@@ -73,6 +74,90 @@ module.exports.sendResponseWithStatusCode = function (res, code = 200, message =
...
+ * @param {string} messageTemplate
+ */
+module.exports.sendReplyWithCampaignTemplate = function (req, res, messageTemplate) {
+  const campaign = req.campaign;
+  let messageText;
+
+  // This virtualProperty check will be deprecated once we add new Contentful fields.
+  // @see https://github.com/DoSomething/gambit-conversations/issues/67
+  const virtualProperty = campaign[messageTemplate];
+  if (virtualProperty) {
```
deadlybutter: Nitpicky, could probably make simple if/elses a one liner `const messageText = virtualProperty ? virtualProperty : campaign.templates[messageTemplate]`

lib/middleware/receive-message/campaign-continue.js
```diff
@@ -1,26 +1,17 @@
...
-
-    // Let's prompt User to continue the Signup conversation for their current Campaign.
-    req.reply.template = 'askContinueMessage';
-
     // Topic may be set to random from the last Rivescript reply.
     // We set it to the Campaign topic to listen for confirmedCampaign, declinedCampaign macros.
     req.conversation.setTopic(req.campaign.topic);
 
-    return next();
+    return helpers.sendReplyWithCampaignTemplate(req, res, 'askContinueMessage');
```
deadlybutter: I think constants like this (`askContinueMessage `) should be defined in a file somewhere

lib/middleware/receive-message/campaign-menu.js
```diff
@@ -3,25 +3,23 @@
...
+    let campaign;
     // Find a random Campaign to prompt for Signup.
     // Eventually query Signups to find Campaigns that are new to User, within their interests, etc.
     return Campaigns.findRandomCampaignNotEqualTo(req.conversation.campaignId)
-      .then((campaign) => {
-        req.campaign = campaign;
-        req.conversation.setCampaign(campaign);
-        req.reply.template = 'askSignupMessage';
+      .then((randomCampaign) => {
+        campaign = randomCampaign;
```
deadlybutter: Again, assigning values outside of a promise chain feels weird to me. It's also not clear why `campaign` is used here.

lib/middleware/receive-message/parse-ask-continue-answer.js
```diff
@@ -1,32 +1,30 @@
...
-      return next();
+      return helpers.sendReplyForCampaignSignupMessage(req, res);
     }
 
-    if (req.reply.brain === 'confirmedCampaign') {
-      req.reply.template = 'gambit';
-
-      return next();
+    let template;
+    if (req.rivescriptReplyText === 'declinedCampaign') {
```
deadlybutter: Again this could probably be a ternary

lib/middleware/receive-message/parse-ask-signup-answer.js
```diff
@@ -1,33 +1,31 @@
...
     if (!askedForSignup) {
       return next();
     }
 
-    if (req.reply.brain === 'declinedCampaign') {
-      req.conversation.declineSignup();
-      req.reply.template = 'declinedSignupMessage';
+    if (req.rivescriptReplyText === 'confirmedCampaign') {
+      // Set this to trigger Campaign Doing Menu reply in Gambit Campaigns.
+      req.keyword = 'confirmed';
```
deadlybutter: In addition to defining the template constants that I mentioned before it seems like we should also define keyword constants

app/routes/receive-message.js
aaronschachter: I've updated the comment to hopefully better explain.

app/models/Conversation.js
aaronschachter: Good 👁 Good 👁 ⚾️

lib/helpers.js
aaronschachter: I've updated comments here too, and modified the 2nd promise to use the `req.outboundMessage` we set in the first.

We don't want to call `res.send` without a successful `postMessageToPlatform`. I've updated the comments to hopefully clarify: `conversation.postMessageToPlatform(message)` makes the POST request to the API of our Conversation (Twilio, Slack, Facebook), posting our Message data to deliver it to the end user.

The `res.send` isn't posting a message -- it's returning the inbound and outbound messages we created for the given`req`. This allows developers to use the Gambit Conversations API and handle delivering the reply (our outbound message) to the User -- which is [how the Consolebot script works.](https://github.com/DoSomething/gambit-conversations/blob/d8878c4888dc8e1233a9803f771ea493487b795b/lib/consolebot.js#L42)

aaronschachter:
@deadlybutter Thank you for the thorough review! I've made a few of the cleanup changes, and aimed to clarify the intent of `helpers.sendReply`. Definitely down to chat through it more if there's better suggestions.

I do agree that the hardcoded Campaign template and keywords can be moved into a config file - but I'd prefer to add that in another PR (and also get @rapala61's take on how best to store them)

lib/helpers.js
deadlybutter: Gotcha, that makes sense!

Issue #49, opened 2017-08-09
Implement Stathat


rapala61:
# Spec
TBD

#Relevant tickets
- Pivotal card: https://www.pivotaltracker.com/story/show/149901018

aaronschachter:
Sounds like our ideal is to view and report on our Messages data in Looker. @mirie and I were chatting about adding Dashbot to give the data team a visual on how reporting could go (inbound vs outbound, common templates, etc).

The [latest Dashbot release allows storing `platformJSON`](https://www.dashbot.io/docs/generic/node) -- data specific to platform, so we could potentially log data points like `campaignId`, and `topic`.

Issue #56, opened 2017-08-10
Save broadcastId to Messages, post to Gambit Campaigns


aaronschachter:
For `POST /receive-message`: If the last outbound message received contains a broadcast, set it to `req.broadcastId` to send to Gambit.

Refs https://github.com/DoSomething/gambit-conversations/pull/99/files#r137044590

PR #99, opened 2017-09-01
Save Gambit Campaigns message templates

### LINKED ISSUE #98: Store the Gambit Campaigns reply template instead of gambitCampaigns
Refs #97 -- when we're forwarding inbound messages to Gambit Campaigns for a reply, we're saving the outbound reply template to `gambitCampaigns` because we don't have the message template available in the response.

Once we do have it, store it. This is necessary for tracking messages by template in #49

aaronschachter:
* Fixes #98: Instead of saving our Gambit Campaigns outbound reply messages with template `gambitCampaigns`, save the template now returned in Gambit Campaigns response added in https://github.com/DoSomething/gambit/pull/956
    * This gets our Messages data in good shape to start logging our outbound templates, per #49 

* Refactors `helpers.sendReply` as individual functions per template, a la https://github.com/DoSomething/gambit/pull/926

### DIFF
--- a/lib/gambit-campaigns.js
+++ b/lib/gambit-campaigns.js
@@ -41,9 +41,10 @@ function parseReceiveMessageRequest(req) {
   }
   const data = {
     phone,
-    campaignId: req.campaign._id,
+    campaignId: req.campaign.id,
     text: req.inboundMessageText,
     mediaUrl: req.mediaUrl,
+    broadcastId: req.broadcastId,
   };
   if (req.keyword) {
     data.keyword = req.keyword.toLowerCase();
--- a/lib/gambit-campaigns.js
+++ b/lib/gambit-campaigns.js
@@ -30,9 +30,33 @@ module.exports.getActiveCampaigns = function () {
 };
 
 /**
+ * Parse our incoming Express request for expected Gambit Campaigns format.
+ * @param {object} req
+ * @return {object}
+ */
+function parseReceiveMessageRequest(req) {
+  let phone = req.conversation._id;
+  if (req.conversation.platform === 'sms') {
+    phone = req.conversation.platformUserId;
+  }
+  const data = {
+    phone,
+    campaignId: req.campaign._id,
+    text: req.inboundMessageText,
+    mediaUrl: req.mediaUrl,
+  };
+  if (req.keyword) {
+    data.keyword = req.keyword.toLowerCase();
+  }
+
+  return data;
+}
+
+/**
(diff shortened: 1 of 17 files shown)


aaronschachter:
A todo task I think we need is to save our template values without the `Message` suffix, e.g. `askSignup`, `completedMenu` instead of `askSignupMessage`, `completedMenuMessage`. I'd prefer to make those changes in a new branch once this is merged to keep the PR's easier to review.

lib/gambit-campaigns.js
```diff
@@ -30,9 +30,34 @@ module.exports.getActiveCampaigns = function () {
...
+  let phone = req.conversation._id;
+  if (req.conversation.platform === 'sms') {
+    phone = req.conversation.platformUserId;
+  }
+  const data = {
+    phone,
+    campaignId: req.campaign.id,
+    text: req.inboundMessageText,
+    mediaUrl: req.mediaUrl,
+    broadcastId: req.broadcastId,
```
aaronschachter: This adds posting the `broadcastId` but we still need to set it within `receive-message` (and also when to NOT set it, e.g. User changes Campaigns, we shouldn't pass their `lastBroadcastId` #56 )

rapala61: I was thinking about this. I'm not sure toggling this at the conversation schema level makes the most sense. What about if we add the broadcastId to the imported outgoing message, this way we can always inspect the last outgoing message's broadcastId, if there is one, then send it in the data object to Campaigns. It would dynamically change as the next outgoing message won't have a broadcastId, thus not being included in the next post to Campaigns.

aaronschachter:
I do like that idea --  the intent of storing on Conversation is to handle the edge case when a User responds with something other than a Yes / No to the Signup Broadcast. We'll figure it out soon in #56!

rapala61:
I see, we could keep adding the `broadcastId` property to the outgoing message until the response is Yes or No. It would work like a tag in a way, once we process an acceptable response, we don't add the `broadcastId` to the outgoing message.

aaronschachter:
After having some time to think about this, I'm leaning towards https://github.com/DoSomething/gambit-conversations/pull/99#discussion_r137053163 -- only passing through the broadcast ID on an inbound message when it's a response to an outbound message with a broadcast Id. 

Handling the non-yes/no responses is a nice to have, but shouldn't be a launch requirement, as we don't support this currently in production with our Gambit Campaigns `POST /chatbot` Broadcast integration. Will update #56 accordingly.

cc @justkika

PR #97, opened 2017-08-31
Standardize POST responses

aaronschachter:
* Closes #39  -- updates the responses of our POST Receive, Send, and Import Message endpoints to return all Message documents created.

* Fixes catching ECONNREFUSED Gambit Campaigns errors (#86) - renames `gambitCampaigns.postSignupMessage `  to `gambitCampaigns.postReceiveMessage` for clarity

* Refactors `sendResponseForError` and `sendGenericErrorResponse` functions as single `sendErrorResposne`

Issue #73, opened 2017-08-18
Check for delivery status and handle retry requests for outbound messages


aaronschachter:
We're currently making a POST to the Conversation medium's API to deliver our outbound messages to Users, but there aren't any checks for failure.

## Updated Scope

Retry requests should not create new outbound messages. #51 solves this problem for inbound messages. It also lays out the foundation for handling outbound messages.

## Tasks

- [x] `import-message` route  - PR #122 
- [x] `send-message` route - PR #120 
- [x] `receive-message` route - PR #133

PR #103, opened 2017-09-06
WIP: Handling Blink retries to prevent message dupes

### LINKED ISSUE #51: Inbound, Outbound Message + Blink retries
Once Blink starts retrying Conversations API requests that fail -- we should not create additional inbound/outbound Message documents for each retry.

rapala61:
## Summary
Starts implementation of this feature according to this diagram.

## Tasks
- [x] Adding middleware to parse `retryCount` and `requestId` into a `req.metadata` object
- [x] Save metadata object to message created
- [x] Check if the request is a "retry" request to load an **inbound** message instead of creating a new one
- [x] Check if the request is a "retry" request to load an **outbound** message instead of creating a new one

## Relevant tickets
Fixes #51

### DIFF
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -159,9 +159,25 @@ conversationSchema.methods.getMessagePayload = function (req = {}) {
   };
 };
 
-conversationSchema.methods.loadMessageAndUpdateMetadataByRequestId = function (requestId,
+conversationSchema.methods.loadInboundMessageAndUpdateMetadataByRequestId = function (requestId,
   metadata = {}) {
-  const query = { conversationId: this._id, 'metadata.requestId': requestId };
+  const query = {
+    conversationId: this._id,
+    'metadata.requestId': requestId,
+    direction: 'inbound',
+  };
+  const update = { metadata };
+  const options = { new: true };
+  return Messages.findOneAndUpdate(query, update, options);
+};
+
+conversationSchema.methods.loadOutboundMessageAndUpdateMetadataByRequestId = function (requestId,
+  metadata = {}) {
+  const query = {
+    conversationId: this._id,
+    'metadata.requestId': requestId,
+    direction: { $regex: /^outbound.*/i },
+  };
   const update = { metadata };
   const options = { new: true };
   return Messages.findOneAndUpdate(query, update, options);
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -162,6 +162,8 @@ conversationSchema.methods.createInboundMessage = function (req) {
   const message = this.getMessagePayload();
   message.text = req.inboundMessageText;
   message.direction = 'inbound';
+  // TODO: attachments should be default in the getMessagePayload method, because we will eventually
+  // have attachments for outbound messages too
   message.attachments = req.attachments;
 
   // TODO: Handle platform dependent message properties here
@@ -228,4 +230,3 @@ conversationSchema.methods.postMessageToPlatform = function (outboundMessage) {
 };
 
 module.exports = mongoose.model('conversations', conversationSchema);
-
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -150,16 +150,25 @@ conversationSchema.methods.declineSignup = function () {
   return this.save();
 };
 
-conversationSchema.methods.getMessagePayload = function () {
+conversationSchema.methods.getMessagePayload = function (req = {}) {
   return {
     conversationId: this,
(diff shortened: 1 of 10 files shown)


aaronschachter:
Our worlds are going to collide soon for sending outbound messages, as I'm working on [fixing the bug I introduced ](https://github.com/DoSomething/gambit-conversations/pull/97#discussion_r137354053) in this branch: https://github.com/DoSomething/gambit-conversations/compare/feature/56-broadcast?expand=1. I'm going to be opening a PR for it shortly....

For this branch you're working on -- can we open a PR on the commit for when retries for inbound message is working, with outbound to come in a later PR?

rapala61:
@aaronschachter  Oh crap, just read you after I pushed the outbound message stuff. Take a look at whats here now, if its ok, ill merge and open a new pull request for the work that is left for the other routes.

aaronschachter:
This is definitely merge conflict city 🌇 with #104. I'll make some comments inline, but here are some thoughts: 

* This is a lot to review, I'd prefer to get just inbound retries into `master` first before taking on outbound retries (outbound retries also seems closely related to #73). Outbound retries may also not be a requirement for launch (currently in production, if a POST request to Mobile Commons fails, the reply is gone. Not that this is ideal, but we also want to keep our launch day checklist of "all the things" as small as we can)

* I've been wondering about whether we should avoid creating the inbound message until we know the outbound message is successfully delivered. And whether a `Request` model might make sense to add -- [we have one in Gambit Campaigns ](https://github.com/DoSomething/gambit/blob/4.6.3/app/models/BotRequest.js)we used to use to query for Broadcast responses but I'm not sure if anyone ever runs queries on this (I do not). 

When an inbound message fails, in theory it'd make sense that we don't process any subsequent inbound messages. If we were processing out of order (request A arrives first, keeps failing -- request B arrives also and succeeds while A is retrying -- B could modify our conversation state such that when A does successfully process, it's too late). #104 introduces a `lastOutboundMessage` on the Conversation -- this is essentially our source of truth of knowing how to handle the next inbound message.

app/models/Conversation.js
```diff
@@ -150,18 +150,45 @@ conversationSchema.methods.declineSignup = function () {
   return this.save();
 };
 
-conversationSchema.methods.getMessagePayload = function () {
+conversationSchema.methods.getMessagePayload = function (req = {}) {
```
aaronschachter: This feels a little odd to add when we have `Conversation.getFromReq` and `Conversation.createFromReq` -- and also brings to mind whether our retry status should live on the Conversation instead of a Message (https://github.com/DoSomething/gambit-conversations/pull/103#issuecomment-327683153)

app/models/Conversation.js
```diff
@@ -150,18 +150,45 @@ conversationSchema.methods.declineSignup = function () {
...
+    conversationId: this._id,
+    'metadata.requestId': requestId,
+    direction: 'inbound',
+  };
+  const update = { metadata };
+  const options = { new: true };
+  return Messages.findOneAndUpdate(query, update, options);
+};
+
+conversationSchema.methods.loadOutboundMessageAndUpdateMetadataByRequestId = function (requestId,
```
aaronschachter: This feels conflicting with the `Conversation.lastOutboundMessage` introduced in #104 -- also brings to mind whether `Conversation.currentInboundMessage` or `Conversation.currentRequest` should be something to consider. If an inbound message B arrives and we're still retrying a previous inbound message A, it could make sense to send a response to requeue message B until our Conversation is no longer in a retry state.

rapala61:
@aaronschachter 

Yeah, there are a few things that have to change in this pull request after seeing whats on #104. I can remove all the outbound retry code, that would fix it, plus will kick the can down to v2 the subject of figuring out the best way to implement it (outbound retries), given your thoughts in the last comment. Your second point, about treating incoming requests as transactions was in my mind too, but given the speed at which we are going, that could be an architectural decision for v2 too.

app/models/Conversation.js
rapala61: Both of those methods deal with Conversation creation, `getMessagePayload` is the only method that generates the payload to create a message. Since `metadata` is a message property that all messages would have, this was the best spot to add it.

I do agree that maybe we should move more `Message` related logic to the `Message` model itself, right now its all in 'Conversation'.

rapala61:
Ill wait to get #104 merged to fix the conflicts and remove outbound retry code from this PR.

rapala61:
Closing this PR as it contains code that is no longer compatible with the state of Master. Ill create another with updated changes.

PR #104, opened 2017-09-06
Send Message fix + Conversation.lastOutboundMessage

aaronschachter:
* Fixes bug introduced in #97, where `conversation.postMessageToPlatform` was mistakenly removed from `POST /send-message`

* Refactors `Conversation.lastOutboundTemplate` string as `Conversation.lastOutboundMessage` reference to the outbound `Message` document created. This will be used to inspect whether the last outbound contained a `broadcastId` (#56)  

* Refactors `Conversation.createOutbound` functions to create outbound message but also post the message to platform when necessary (for the `outbound-reply`, `outbound-api-send` directions)

### DIFF
--- a/app/models/Campaign.js
+++ b/app/models/Campaign.js
@@ -168,4 +168,4 @@ campaignSchema.virtual('isClosed').get(function () {
 
 /* eslint-enable prefer-arrow-callback */
 
-module.exports = mongoose.model('campaigns', campaignSchema);
+module.exports = mongoose.model('Campaign', campaignSchema);
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -231,7 +231,7 @@ conversationSchema.methods.outboundSend = function (text, template) {
  * @param {string} template
  * @return {Promise}
  */
-conversationSchema.methods.createOutboundImportMessage = function (text, template) {
+conversationSchema.methods.outboundImport = function (text, template) {
   return this.createOutboundMessage(text, template, 'outbound-api-import');
 };
 
@@ -265,4 +265,3 @@ conversationSchema.methods.postMessageToPlatform = function () {
 };
 
 module.exports = mongoose.model('Conversation', conversationSchema);
-
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -23,7 +23,10 @@ const conversationSchema = new mongoose.Schema({
   topic: String,
   campaignId: Number,
   signupStatus: String,
-  lastOutboundTemplate: String,
+  lastOutboundMessage: {
+    type: mongoose.Schema.Types.ObjectId,
+    ref: 'Message',
+  },
   slackChannel: String,
   lastBroadcastId: String,
 }, { timestamps: true });
@@ -58,7 +61,7 @@ conversationSchema.statics.getFromReq = function (req) {
   };
   logger.trace('Conversation.getFromReq', query);
 
-  return this.findOne(query);
+  return this.findOne(query).populate('lastOutboundMessage');
 };
 
 /**
@@ -90,8 +93,6 @@ conversationSchema.methods.setTopic = function (newTopic) {
  * Set topic to random to upause User.
  */
 conversationSchema.methods.supportResolved = function () {
-  this.lastOutboundTemplate = 'front';
-
   return this.setTopic(defaultTopic);
 };
 
@@ -150,12 +151,25 @@ conversationSchema.methods.declineSignup = function () {
   return this.save();
 };
 
-conversationSchema.methods.getMessagePayload = function () {
-  return {
+/**
+ * Gets data for a Conversation Message.
+ * @param {string} text
+ * @param {string} template
+ * @return {object}
+ */
+conversationSchema.methods.getMessagePayload = function (text, template) {
+  const data = {
     conversationId: this,
     campaignId: this.campaignId,
     topic: this.topic,
   };
+  if (text) {
+    data.text = text;
+  }
+  if (template) {
+    data.template = template;
+  }
+  return data;
 };
 
 conversationSchema.methods.createInboundMessage = function (req) {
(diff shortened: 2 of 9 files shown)


aaronschachter:
Deployed to staging, verified that `POST /send-message` works when sending a template value like `externalSignupMenu`, also confirmed messages from Front are working:

screen shot 2017-09-06 at 10 42 19 pm

app/models/Campaign.js
```diff
@@ -168,4 +168,4 @@ campaignSchema.virtual('isClosed').get(function () {
 
 /* eslint-enable prefer-arrow-callback */
 
-module.exports = mongoose.model('campaigns', campaignSchema);
+module.exports = mongoose.model('Campaign', campaignSchema);
```
aaronschachter: Per http://mongoosejs.com/docs/models.html -- first argument should be singular.

app/routes/index.js
```diff
@@ -1,16 +1,36 @@
...
 
+const router = express.Router();
+
+const ConversationModel = require('../models/Conversation');
+const MessageModel = require('../models/Message');
+const CampaignModel = require('../models/Campaign');
+
+restify.serve(router, ConversationModel, { name: 'conversations' });
+restify.serve(router, MessageModel, { name: 'messages' });
+restify.serve(router, CampaignModel, { name: 'campaigns' });
```
aaronschachter: This change https://github.com/DoSomething/gambit-conversations/pull/104/files#r137450464 requires passing a `name` parameter to avoid defining our endpoints with the singular namespace (e.g. we don't want ` GET /api/v1/campaign`)

app/models/Conversation.js
```diff
@@ -170,47 +182,69 @@ conversationSchema.methods.createInboundMessage = function (req) {
...
+ * @param {string} template
+ * @param {string} direction
+ * @return {Promise}
  */
-conversationSchema.methods.createOutboundReplyMessage = function (messageText, messageTemplate) {
-  const message = this.getMessagePayload();
-  message.text = messageText;
-  message.template = messageTemplate;
-  message.direction = 'outbound-reply';
+conversationSchema.methods.createOutboundMessage = function (text, template, direction) {
```
rapala61: Since we are also sending direction, should the method be named with a static `outbound` direction?

app/models/Conversation.js
```diff
@@ -170,47 +182,69 @@ conversationSchema.methods.createInboundMessage = function (req) {
...
-  const message = this.getMessagePayload();
-  message.text = messageText;
-  message.template = messageTemplate;
-  message.direction = 'outbound-api-import';
+/**
+ * @param {string} text
+ * @param {string} template
+ * @return {Promise}
+ */
+conversationSchema.methods.outboundReply = function (text, template) {
```
rapala61: In the same context of [above comment](https://github.com/DoSomething/gambit-conversations/pull/104/files#r137572602), could the name be a little more specific about what action is this method performing, like `createOutboundReply` VS `outboundReply`?

app/models/Message.js
```diff
@@ -22,4 +22,4 @@ const messageSchema = new mongoose.Schema({
   attachments: Array,
 }, { timestamps: true });
 
-module.exports = mongoose.model('messages', messageSchema);
+module.exports = mongoose.model('Message', messageSchema);
```
rapala61: Nice, 👍

app/models/Conversation.js
aaronschachter: I'm not sure what you mean, could give you an example?

app/models/Conversation.js
aaronschachter: I did drop the `create` part from the name here because we're posting the message to the Conversation platform as well as creating the outbound message. Maybe starting it with the verb would read better? e.g. `conversation.replyOutbound(text, template)`, `sendOutbound`, `importOutbound`

app/models/Conversation.js
rapala61: I'm just wondering if this method could also be used to create an inbound message, since it also accepts a direction argument. It seems to me like sending a direction dependency makes it a little confusing to use a deterministic name like `createOutboundMessage`.

aaronschachter: Gotcha, that checks out. I'm into it - we'd just need to change the Receive Message middleware that calls `createInboundMessage` to call `createMessage(req.inboundMessageText, null, 'inbound')` instead. I'll make the change?

aaronschachter: I forgot that `createOutboundMessage` is also setting the `Conversation.lastOutboundMessage`, as a means to DRY in the `outboundReply`, `outboundSend`, and `outboundImport` methods.

 `createInboundMessage` is also setting `req.attachments`.

So potentially it could go something like:
* `Conversation.createMessage(direction, text, template, attachments)`
* `Conversation.createLastOutboundMessage(direction, text, template, attachments)` that calls `this.createMessage` and then sets `this.lastOutboundMessage`

rapala61: The two functions sound like the way to go. About the attachments property, I think its part of other properties that live in the `req` object that I think makes more sense to include in the `getMessagePayload` method. This is why I starting sending the `req` object to this method. Otherwise, we will have to do the same with metadata, include it as an argument to `Conversation.createMessage` and `Conversation.createLastOutboundMessage` methods respectively.

app/models/Conversation.js
rapala61: This is a hard one, ideally, I think we would want to have the method have one responsibility only, otherwise the name is vague because we can't determine with exactitude what its doing, in this case is creating AND posting, but not with import, because import messages don't need to be posted, just imported. What I have seen before to fix this is creating another function named something like `createAndPost`, which would ideally have both calls to `createOutboundReply` and `postOutbound`.

app/models/Conversation.js
aaronschachter: Passing `req.attachments` is where I get a little lost on passing `req` and having one `createMessage` function. What if there's attachments for both inbound and outbound messages? 
Let's use those 2 functions as defined in the comments above, passing `attachments` as a parameter for now to unblock #103 and we can revisit using `req`when Inbound Retries are ready.

aaronschachter: Looking into adding the `Message.broadcastId` in #56 and see what you mean. Going to hold on adding it until #103 is ready for review to avoid more collisions 🔥

PR #105, opened 2017-09-07
Handling Blink retries to prevent inbound message dupes

### LINKED ISSUE #51: Inbound, Outbound Message + Blink retries

rapala61:
## Summary
Starts implementation of this feature according to this diagram.

## Tasks
- [x] Adding middleware to parse `retryCount` and `requestId` into a `req.metadata` object
- [x] Save metadata object to message created
- [x] Check if the request is a "retry" request to load an **inbound** message instead of creating a new one

## Relevant tickets
Fixes #51

### DIFF
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -150,12 +150,35 @@ conversationSchema.methods.declineSignup = function () {
 };
 
 /**
+ * Loads the inbound message that matches this conversationId and
+ * the metadata.requestId and updates its metadata with the new one.
+ * @param {number} requestId
+ * @param {object} metadata
+ * @return {object}
+ */
+conversationSchema.methods.loadInboundMessageAndUpdateMetadataByRequestId = function (requestId,
+  metadata = {}) {
+  const query = {
+    conversationId: this._id,
+    'metadata.requestId': requestId,
+    direction: 'inbound',
+  };
+  const update = { metadata };
+  const options = { new: true };
+  return Messages.findOneAndUpdate(query, update, options)
+    .then(message => message.populate('conversationId').execPopulate())
+    .then(() => {
+      this.populate('lastOutboundMessage').execPopulate();
+    });
+};
+
+/**
  * Gets data for a Conversation Message.
  * @param {string} text
  * @param {string} template
  * @return {object}
  */
-conversationSchema.methods.getMessagePayload = function (text, template) {
+conversationSchema.methods.getDefaultMessagePayload = function (text, template) {
   const data = {
     conversationId: this,
     campaignId: this.campaignId,
@@ -171,6 +194,22 @@ conversationSchema.methods.getMessagePayload = function (text, template) {
 };
 
 /**
+ * Gets data from a req object for a Conversation Message.
+ * @param {string} text
+ * @param {string} template
+ * @return {object}
+ */
+conversationSchema.methods.getMessagePayloadFromReq = function (req = {}) {
+  // TODO: Handle platform dependent message properties here
+  const data = {
+    metadata: req.metadata || {},
+    attachments: req.attachments || [],
+  };
+  return data;
+};
+
+
+/**
  * Creates Message for a Conversation with given params.
  * @param {string} direction
  * @param {string} text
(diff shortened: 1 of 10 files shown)


rapala61:
Tested on staging through slack. Tested locally by posting to `import-message`, `send-message`, and `receive-message`. All messages included metadata. Inbound messages are loaded instead of created when `retryCount` > 0

app/models/Conversation.js
```diff
@@ -150,12 +150,34 @@ conversationSchema.methods.declineSignup = function () {
...
+  metadata = {}) {
+  const query = {
+    conversationId: this._id,
+    'metadata.requestId': requestId,
+    direction: 'inbound',
+  };
+  const update = { metadata };
+  const options = { new: true };
+
+  return this.populate('lastOutboundMessage').execPopulate()
```
aaronschachter: Also not sure why we're populating `lastOutboundMessage` here -- [this should be populated already](url) by the time we call this function?

lib/middleware/receive-message/message-inbound-load.js
```diff
@@ -0,0 +1,26 @@
...
+    if (!req.isARetryRequest()) {
+      return next();
+    }
+
+    // TODO: This should be moved to receive-message/params middleware
+    if (req.inboundMessageText) {
+      req.userCommand = req.inboundMessageText.trim();
+    }
+
+    return req.conversation.loadInboundMessageAndUpdateMetadataByRequestId(req.metadata.requestId,
```
aaronschachter: Per comments on this function -- could it make sense to define this as a static method in our `Messages` model that accepts an additional `conversationId` ?

app/models/Conversation.js
```diff
@@ -150,12 +150,34 @@ conversationSchema.methods.declineSignup = function () {
...
+    .then(() => Messages.findOneAndUpdate(query, update, options)
+      .then(message => message.populate('conversationId').execPopulate()));
```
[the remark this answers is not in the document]
rapala61: The `Message.create` method populates the `conversationId` property with the current conversation instance by default. The `Message.findOneAndUpdate` doesn't return the same populated result so I had to populate it manually, to keep it consistent with the response when the inbound message is created instead of loaded.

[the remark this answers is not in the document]
aaronschachter: Ah. I've been trying to figure out how it the inboudn Message.conversation gets loaded at all - I think ideally it shouldn't be in the `data.inbound` message.
I've been considering whether we should add the Conversation as a property of data for all of our `POST` requests... so for Receive message something liek:
```
data: {
   conversation: {
   }
   messages: {
     inbound: [],
     outbound: [], // We only return one now, but ideally we should return multiple for those long paragraph Signup menus etc.
   }
}
```

app/models/Conversation.js
rapala61: Same here. Even after manually populating the `conversationId`, the `lastOutboundMessage` did not get populated in the loaded message.

app/models/Conversation.js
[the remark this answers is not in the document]
rapala61: I see. It threw me off that you were explicitly populating the `lastOutboundMessage`. It made it seem like having the conversation populated in the inbound was important. I can get rid of that manual population when loading, since this is the desired behavior.

lib/middleware/receive-message/message-inbound-load.js
rapala61: I can make `loadInboundMessageAndUpdateMetadataByRequestId` a static method. About the other method, would it retrieve all messages for given campaign? like `loadInboundMessagesByConversationId `?

aaronschachter: I was thinking one method -- instead of `Conversation. loadInboundMessageAndUpdateMetadataByRequestId`

it's `Messages.getInboundMessageByRequestId`, I guess we wouldn't really need the conversation ID at all since it's already stored on the Message and we'd only have one per Request ID

app/models/Conversation.js
```diff
@@ -170,25 +170,40 @@ conversationSchema.methods.getMessagePayload = function (text, template) {
...
+ * Gets data from a req object for a Conversation Message.
+ * @param {string} text
+ * @param {string} template
+ * @return {object}
+ */
+conversationSchema.methods.getMessagePayloadFromReq = function (req = {}) {
+  // TODO: Handle platform dependent message properties here
+  const data = {
+    metadata: req.metadata || {},
+    attachments: req.attachments || [],
```
aaronschachter: Refs #109

PR #113, opened 2017-09-11
WIP: Adds `platformMessageId` to Messages

### LINKED ISSUE #52: Store external Message ID's in our Messages
We'll want a property named something like `mediumMessageId` that corresponds to the Twilio, Messenger, or Slack ID for both an inbound message, and the outbound messages we create and post to the respective platform API.

## Updated scope

Property name will be `platformMessageId`.

Refs #39

rapala61:
## Summary
It will parse the `platformMessageId` from incoming requests and add it to created messages.
It also includes some refactoring and some new helpers

## Related issues
Fixes #52


rapala61:
@aaronschachter What would be the `platform` for an outbound front support message? I'm noticing we are not setting it. I wonder if it should be `api`, like for receiving generic messages.

lib/middleware/import-message/broadcast.js
```diff
@@ -43,10 +49,14 @@ module.exports = function getBroadcast() {
   return (req, res, next) => {
     const asyncProperties = [];
 
+    // We are only allowing Customer.io imports, which MUST have broadcastId. If in the future
+    // we want to allow generic imports we can relax this requirement here and call next()
     if (!req.broadcastId) {
-      return next();
+      const error = new UnprocessibleEntityError('requestId is a required property.');
```
aaronschachter: Typo? Should be `broadcastId`, not `requestId`

lib/helpers.js
```diff
@@ -1,11 +1,209 @@
...
+    req.platformUserId = req.body.phone;
+    req.broadcastId = req.body.broadcastId;
+    req.messageFields = req.body.fields;
+  },
+};
+
+/**
+ * Twilio helper
+ */
+module.exports.twilio = {
```
aaronschachter: Question more for discussion vs change request, but would it make sense to define these in `/lib/twilio` instead, to potentially avoid an unwieldy `/lib/helpers`?

lib/helpers.js
```diff
@@ -1,11 +1,209 @@
...
+    return !!req.body.slackId;
+  },
+  isFacebook: function isFacebook(req) {
+    return !!req.body.facebookId;
+  },
+  isFront: function isFront(req) {
+    // TODO: Should be in a config constant
+    return !!req.get('x-front-signature');
+  },
+  isCustomerIO: function isCustomerIO(req) {
```
aaronschachter: This naming convention is inconsistent with `helpers.customerIo`. Let's pick one and stick with it?

lib/middleware/import-message/broadcast.js
```diff
@@ -3,7 +3,9 @@
 const Promise = require('bluebird');
 
 const helpers = require('../../helpers');
+const UnprocessibleEntityError = require('../../../app/exceptions/UnprocessibleEntityError');
 
+// TODO: Move inside helpers.customerIo
```
aaronschachter: Wouldn't these fall under a `helpers.contentful` (or `/lib/contentful` as per previous comment) instead of Customer IO?

lib/helpers.js
rapala61: Yeah, I can better organize it by having separate helpers inside a `lib/helpers` folder: `lib/helpers/twilio`, `lib/helpers/attachments`, etc. Then require them all inside a `helpers/index.js`. I would like them all to exist in a helpers module, although organized separately.

rapala61: @aaronschachter Is it ok to make a soft move towards this new org by just starting with the new helpers? I don't want to fully migrate the hole helpers module and make this PR even bigger.

rapala61:
Updated.

PR #120, opened 2017-09-13
Handling Blink retries to prevent outbound message dupes in `send-message` route

rapala61:
## Summary
Checks if the request contains retry count headers . If it does, it loads the outbound message based on its `metadata.requestId` and direction, instead of creating a new one, preventing duplicates.

## Related issues
Issue #73 
Part B of #105

PR #48, opened 2017-08-09
Import message

rapala61:
## Summary
This PR implements the `import-message` endpoint. This endpoint is intended to import a message for a given user conversation without sending it to the user itself. As of now, the only medium accepted is `customerio`. I'm creating the PR for testing and review purposes, its not to be merged yet, until everyone feels like is the right time and its ready.

## Testing
You can test this in 2 ways:

1. By sending a webhook request via Customer.io - Complicated if you are not familiar with the steps, which include looking at Contentful, Customer.io, ngrok, etc.

2. Using Postman, sent a POST request to your running Convo API: `http://puppet:totallysecret@<your localhost>/api/v1/import-message?medium=customerio`.
    - The `Content-Type` should be `application/json` and the payload should be: 
```
{
	"broadcast_id": "7zU0Mb1k9GkWWI40o06Mic",
	"phone": "<THE PHONE NUMBER, EXP: +5556667788>",
	"fields": [
		{
			"customer.first_name": "<YOUR NAME>"
		}
	]
}
```

## Success
- If a conversation for this user exists, the conversation should be updated and a new outbound message should have been created, as per this specs: #34 

## Relevant Ticket
- Fixes #34

PR #122, opened 2017-09-13
Handling Blink retries to prevent outbound message dupes in `import-message` route

rapala61:
## Summary

## Related issues
Issue #73 
Part B of #105


lib/middleware/import-message/message-outbound-load.js
```diff
@@ -0,0 +1,29 @@
...
+const Message = require('../../../app/models/Message');
+
+module.exports = function loadOutboundImportMessage() {
+  return (req, res, next) => {
+    // If this is not a retry request, we have to create a new outbound message
+    if (!req.isARetryRequest()) {
+      return next();
+    }
+    logger.debug('loadOutboundImportMessage: Is a retry request. Proceeding with loading message.');
+    return Message.updateOutboundApiImportMessageMetadataByRequestId(req.metadata.requestId,
```
aaronschachter: This middleware file feels like it's essentially identical to one in [Receive Message](https://github.com/DoSomething/gambit-conversations/blob/5a742cad433b6c6ad3d1829f284793923922d823/lib/middleware/receive-message/message-inbound-load.js#L18) except for the `direction` value we save on the outbound message.

Would it be cleaner to define `message-outbound-load` and `message-outbound-create` in `/lib/middleware`, where this function expects something like a `req.outboundMessageDirection`? Because we're passing both the `req.metadata.requestId` and `req.metadata` -- maybe it just needs to accept a single `req`? e.g. `Message.updateOutboundMessageByRequest(req)`

rapala61: Both middleware load a message, but aside from being very clear about which message is loading, there is one major difference in the `send-message/message-outbound-load.js`. It calls `req.conversation.postLastOutboundMessageToPlatform()`.

lib/middleware/import-message/message-outbound-load.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+const logger = require('heroku-logger');
+
+const helpers = require('../../helpers');
+const Message = require('../../../app/models/Message');
```
aaronschachter: This is https://github.com/DoSomething/gambit-conversations/pull/48#discussion_r132507733 -- let's keep naming consistent. Again, I'm fine with renaming our Model `const` declarations to singular but I'd prefer we keep consistent int he files.

lib/middleware/import-message/message-outbound-load.js
aaronschachter: You mean Receive Message, right? Send Message will need the same -- posting message to platform. One idea that comes to mind that could potentially be cleaner could be a config file for general `message-outbound-load/send` to specify the `direction` and a `boolean` whether to postToPlatform?

rapala61: no, I mean `send-message`. It's in the pull request you reviewed earlier. https://github.com/DoSomething/gambit-conversations/pull/120/files#diff-e64e0deb429fb33a8a363313e4449178R26

lib/middleware/import-message/message-outbound-load.js
rapala61: I don't understand. You actually introduced the change to start naming the model by its singular here: https://github.com/DoSomething/gambit-conversations/pull/104/files#diff-adc1cae48edeabf707d9aa9f1bc46920R25

rapala61: @aaronschachter I followed the change you made [here](https://github.com/DoSomething/gambit-conversations/pull/104/files#diff-adc1cae48edeabf707d9aa9f1bc46920R25), that's why I used singular. If you think I have to change it to plural, I'll do it, because I want to merge this as soon as possible.

lib/middleware/import-message/message-outbound-load.js
rapala61: This PR adds the `message-outbound-load` to the `import-message` route. The previous PR #120, added the same middleware, but for `send-message`. The `receive-message` route doesn't use this pattern to send reply messages. It uses the helpers module. I will have to intercept the `sendReply` method there to check if the request is a retry request and load the message instead.

lib/middleware/import-message/message-outbound-load.js
aaronschachter: I take it you're referring to `module.exports = mongoose.model('Message', messageSchema);`? That 1st parameter change for `mongoose.model` is the singular so setting the `schema.lastOutboundMessage.ref` to `Message` works in #104. 

This is same comment as https://github.com/DoSomething/gambit-conversations/pull/48#discussion_r132507733 -- all middleware has been defining models as `const Campaigns = require('./models/Campaigns')`. It's a little bit of a nitpick -- but it's like defining `loggers` instead of `logger` in a middleware when we use `logger` everywhere. And totally fine with renaming them all as singular, I just think our naming conventions should be consistent when possible.

rapala61: Yes, that is what I was referring to.  I get the inconsistency argument, I agree with you. What I would like for all of us to discuss as a team in a Dev Jam, is consistency across the board, in this case is naming, in others is PR size, etc. For example: https://github.com/DoSomething/gambit-conversations/blob/master/lib/middleware/conversation-create.js#L3. This is a singular definition of the `Campaign` model, which is inconsistent with other parts of the code and with your example above.

I want to fix this, do you think that if I go through the whole codebase and change all instances of models to singular, we can move on with this PR?

aaronschachter: Let's change the middleware in a different branch -- didn't realize Conversation Create was inconsistent. Happy to chat through those topics in Dev Jam.

aaronschachter:
I'm still interested in whether single Outbound Message Load/Create middleware would DRY things quite a bit - but not necessary for merge...

rapala61:
Thanks. Will DRY it up in the next PR!

Issue #63, opened 2017-08-15
Account preferences Conversation topic


aaronschachter:
A `LESS` keyword should trigger the following:
* [ ] **BLOCKED** -- Update the User's Northstar profile, setting their preferences to receive SMS Signup Broadcasts as Monthly
* [ ] Respond with a message that we've updated their preferences

Some other keyword should do the same, except set their preferences to receive the broadcasts more often -- weekly? Need spec for this. 
cc: @justkika

justkika:
on it! I need to find out the current rules for how people get into these groups and I'll document.

aaronschachter:
Updated this description to include opt-outs (STOP)

aaronschachter:
When a User texts STOP -- do they need to text a magic keyword to join again? Or do we accept any new inbound message as a re-subscription?

@justkika @classicfreddie

aaronschachter:
We chatted through this today -- for v1, we'll accept any inbound message and respond with the Gambit reply (e.g. User doesn't need to text START or JOIN to re-subscribe).

Need to jam with @DFurnes to spec how to store this status in Northstar.

DFurnes:
Chatted this over [with Kika this morning](https://dosomething.slack.com/archives/C2C8NLNAY/p1505401949000378). We're going to rename the existing `mobilecommons_status` field in Northstar to `mobile_status` and add an additional `less` state there to track users who are subscribed but want less messaging.

I'll also add a `last_messaged_at` field for tracking the last time a user sent a message to us, similar to how we track `last_accessed_at` and `last_authenticated_at` for web traffic.

PR #130, opened 2017-09-14
Get User middleware

aaronschachter:
Another step towards #63: Adds Get User middleware to Receive Message for SMS Conversations, sets a `req.user` if found.


rapala61:
I feel like this step here has a good potential of slowing down the Convo API. I'm don't know the entire scope of this, and maybe you already figured out, but I suggest we just send this request to blink in the future and always respond with success to the user. Even though updating subscription is important, I don't think is critical enough to be able to break the flow of an user's experience.

aaronschachter:
Agree - always responding with success is ideal!

PR #132, opened 2017-09-15
DRY message-outbound-load and message-outbound-create

rapala61:
## Summary
It refactors the middleware that loads outbound messages on retries for `send-message` and `import-message` routes. It also goes a step further and refactors the outbound message creation for the same routes. Some minor bugs fixed on along the way.

## This is a medium size PR
It contains only the refactoring of the already implemented middleware for those routes: `send-message` and `import-message`. The actual handling of retries on whats left of #73 will be included on a different PR.

Based on @aaronschachter and I conversation here: https://github.com/DoSomething/gambit-conversations/pull/122#issuecomment-329539010.


rapala61:
Made the changes!

PR #133, opened 2017-09-15
Handling Blink retries to prevent outbound message dupes in `receive-message` route

### LINKED ISSUE #73: Check for delivery status and handle retry requests for outbound messages

rapala61:
## Summary

## Related issues
Fixes #73 
Part B of #105 

## Test
You can test with Consolebot. Just type `retry` and it will attempt to retry the last message sent to Conversations.


lib/consolebot.js
```diff
@@ -26,18 +31,37 @@ class Consolebot {
...
+    if (data.headers && data.headers[retryHeader]) {
+      payload = data;
+    } else {
+      const defaultPayload = {
+        platformUserId: this.options.platformUserId,
+        text: '',
+        headers: {},
+      };
+      defaultPayload.headers[retryHeader] = 0;
+      defaultPayload.headers[requestIdHeader] = uuidv4();
```
aaronschachter: Just curious - is there a reason we need a `UUID` and the module vs just using something random like the current timestamp?

package.json
```diff
@@ -44,7 +44,8 @@
     "rivescript": "^1.17.2",
     "superagent": "^3.5.2",
     "twilio": "^3.5.0",
-    "underscore": "^1.8.3"
+    "underscore": "^1.8.3",
+    "uuid": "^3.1.0"
```
aaronschachter: Should this be defined in `devDependencies` since Consolebot is the only place we use it? It's where we've defined `readline` and `nock`...

lib/consolebot.js
rapala61: I read somewhere in [here](https://devcenter.heroku.com/articles/http-request-id) that UUID seems to be the standard for request ids. It also says that it can be any combination of alphanumeric characters and some symbols, between 20 and 200 characters. Using the UUID package is not a requirement.

package.json
rapala61: I'm seeing a lint error now that I moved this dependency into devDependencies. I'll move back to dependencies as this package is technically used in a script and not in a test.

```
/Users/rpacas/Desktop/repos/gc/lib/consolebot.js
  8:16  error  'uuid' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
```

Issue #139, opened 2017-09-21
Sending SMS messages through Twilio and Customer.io, using Webhooks instead of Twilio integration.


rapala61:
# Context
We are currently using Customer.io to create segmented email campaigns using it's email integration. We also need to support triggered campaigns that send SMS messages through Twilio. Since SMS is a central part of how users interact with us, we need to integrate (import) these messages sent from Customer.io into each user's message history if the user has successfully received it.

Our DoSomething SMS platform (Gambit) must be aware that the user has been sent a message, which means we must update the user state in our platform **before** we are ready to process the user's response via SMS. This flow creates a technical challenge for us. How do we assure that the user will have it's state updated successfully every time on our SMS platform, only for those users who do get the message to their devices?

# Solutions
## Backup solution (DEPRECATED)
Our current plan is based on using both the `Webhook` and the `Twilio SMS` workflow steps available in Customer.io. The `Webhook` would be used to update the user's state. The `Twilio SMS` integration would be used to send the text message to the user. The Webhook step would be send automatically, while the Twilio SMS step would be drafted and only sent after all webhook requests have been processed.

### Trade offs
- (+) We make use of the tools already in Customer.io.
- (+) We can guarantee that all users have their state (in broadcast) updated prior to sending them the SMS message.
- (-) We have to manually execute the step of sending the SMS messages.
- (-) The change in the user's state (Webhook) is blind to the outcome of the Twilio SMS message. We would have to blindly trust that the user got the text.
- (-) The longer the list of recipients in this campaign. The longer the gap will be between processing Webhooks by our platform and users getting their SMS message from Customer.io. Potentially allowing users to communicate with us via SMS without knowing they are in broadcast mode, because their state has been changed, but they have not received the SMS text.

## Proof-of-concept Solution (Implemented in #141 #151)
The POC solution doesn't use the Twilio SMS integration in Customer.io to send messages to the users. Instead, it uses the raw Twilio's REST Messaging API by leveraging the Webhook workflow step in Customer.io. Twilio's API is very flexible and allows us to setup of a `statusCallback` parameter which acts as its own webhook that would give us updates on the status of the sent SMS text (sent, delivered, failed, etc). Our platform would process these events. We would update **only** the state of users who have successfully received the SMS message to their device.

[...]

Issue #123, opened 2017-09-14
Create documentation on how to create a broadcast


rapala61:
## Summary
The creation of a broadcast process should be documented in a way that can be followed by all DoSomething staff.

## Tasks
- [x] Update documentation for the `import-message` route. - PR #143 
- [x] Update the `Broadcasts` Wiki page with the new process diagram.
- [x] Create step-by-step guide in the Wiki to create and execute a new broadcast. - [Broadcasts wiki](https://github.com/DoSomething/gambit-conversations/wiki/Broadcasts)

## Blocks
Blocked by the `broadcast-settings` route implementation in #142. - Unblocked by #151

rapala61:
![broadcasts v2 - page 1](https://user-images.githubusercontent.com/863301/30979066-2028c068-a442-11e7-87f9-207b138e5661.png)

rapala61:
![broadcasts v2 - page 1 1](https://user-images.githubusercontent.com/863301/30979503-6db05e44-a443-11e7-85a7-f5f13c77ac63.png)

rapala61:
<img width="419" alt="screen shot 2017-10-03 at 11 09 55 am" src="https://user-images.githubusercontent.com/863301/31136132-e3abf3e6-a82c-11e7-826d-e72ea17258a6.png">

rapala61:
<img width="451" alt="screen shot 2017-10-03 at 3 14 07 pm" src="https://user-images.githubusercontent.com/863301/31147377-533d0d5e-a84f-11e7-9fe5-5a68440db65c.png">

<img width="363" alt="screen shot 2017-10-03 at 3 14 49 pm" src="https://user-images.githubusercontent.com/863301/31147392-621d186e-a84f-11e7-8b84-757892f8b827.png">

<img width="479" alt="screen shot 2017-10-03 at 3 15 19 pm" src="https://user-images.githubusercontent.com/863301/31147401-68cf2fda-a84f-11e7-8eb4-dc8933365a13.png">

<img width="572" alt="screen shot 2017-10-03 at 3 15 43 pm" src="https://user-images.githubusercontent.com/863301/31147414-793a8f54-a84f-11e7-8a0a-cb174b47a672.png">

<img width="438" alt="screen shot 2017-10-03 at 3 16 04 pm" src="https://user-images.githubusercontent.com/863301/31147427-87af26a8-a84f-11e7-8e09-d41b8ce4fa1d.png">

[...]

PR #141, opened 2017-09-26
Import of broadcast message through Twilio statusCallback POST requests

rapala61:
# Summary
This is the code that supports the Proof-Of-Concept approach to broadcasts depicted in #139.

# Things to consider
- **Twilio**
    - Twilio will POST (n) requests to the URL we set up as `statusCallback` URI. The amount of requests depend on many factors, including the carrier. A list of possible statuses sent to the `statusCallback` URI can be [found here](https://www.twilio.com/docs/api/messaging/message#message-status-values).
        - Example: My carrier is T-mobile. The statuses Twilio POSTed to my local Convo API while developing were: `queued`, `sent`, and `delivered`.
    - We are only importing messages that have been `delivered` to the user. This filtering should happen at the Blink worker level so we don't run the risk of overloading the Convo API. However, is important to respond with success for all message statuses, even if we are filtering them out, otherwise we will get [#11200 errors](https://www.twilio.com/docs/api/errors/11200) in our Twilio dashboard.
- **Security**
    - We are using a Twilio revokable API Key as an added layer of security when creating messages from the Customer.io's Webhook workflow step.
    - Twilio requests should be validated against their `X-Twilio-Signature` header. This check should happen at the Blink worker level, since it gets the request directly from Twilio. For testing, I have enabled API key authentication, so that the `statusCallback` URI can contain an API Key as a query parameter and be authenticated.
        - [Docs about how to do this in Node and Express using the webhook() middleware](https://www.twilio.com/docs/guides/how-to-secure-your-express-app-by-validating-incoming-twilio-requests#use-twilio-express-request-validation-middleware)
        - [API Reference docs on the methods we can use to do this manually](https://twilio.github.io/twilio-node/3.5.0/global.html#validateExpressRequest)
    - ~~Basic Auth in the `statusCallback` doesn't seem to be supported.~~ [It is.](https://github.com/DoSomething/gambit-conversations/pull/141#issuecomment-332318577)
- **Customer.io**
    - The Webhook workflow step must be setup with `application/x-www-form-urlencoded` content type headers. `application/json` [is NOT supported](https://www.twilio.com/docs/api/rest/request#post).
    - The `statusCallback` URI sent in the body of the webhook **must** be [URI encoded](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent) in order to preserve its query parameters like `broadcastId`, and `apiKey` (for now).
        - Example of the Webhook body:
        ```To={{customer.phone}}&MessagingServiceSid=secretSauce&Body=tacos are great&StatusCallback=http%3A%2F%2F57528dc6.ngrok.io%2Fapi%2Fv1%2Fimport-message%3FbroadcastId%3Dtacosfest%26apiKey%3Dthe-api-key-yo```

# Next tasks to work on
- Add route to generate the Webhook body, given a `broadcastId` and `apiKey`.
- Add documentation for this broadcast process #123.
- Remove code that supports the `Backup solution` in #139.

# Relevant Issues
Ref #139 
Ref #123

### DIFF
--- a/lib/helpers/broadcast.js
+++ b/lib/helpers/broadcast.js
@@ -1,7 +1,11 @@
 'use strict';
 
+function getBroadcastId(req) {
+  return req.body.broadcastId || req.query.broadcastId;
+}
+
 /**
- * Customer.io helper
+ * Broadcast helper
  */
 module.exports = {
   /**
@@ -11,8 +15,6 @@ module.exports = {
    * @return {promise}
    */
   parseBody: function parseBody(req) {
-    req.platform = 'sms';
-    req.platformUserId = req.body.phone;
-    req.messageFields = req.body.fields;
+    req.broadcastId = getBroadcastId(req);
   },
 };
--- a/lib/helpers/broadcast.js
+++ b/lib/helpers/broadcast.js
@@ -9,7 +9,7 @@ function getBroadcastId(req) {
  */
 module.exports = {
   /**
-   * parseBody - parses properties out of the body of a Facebook request
+   * parseBody
    *
    * @param  {object} req
    * @return {promise}
--- a/lib/helpers/request.js
+++ b/lib/helpers/request.js
@@ -22,7 +22,7 @@ module.exports = {
     // TODO: Should be in a config constant
     return platform === 'customerio';
   },
-  isStatusCallback: function isStatusCallback(req) {
+  isTwilioStatusCallback: function isTwilioStatusCallback(req) {
     const messageStatus = req.body.MessageStatus;
 
     /**
--- a/lib/helpers/request.js
+++ b/lib/helpers/request.js
@@ -22,4 +22,14 @@ module.exports = {
     // TODO: Should be in a config constant
     return platform === 'customerio';
   },
+  isStatusCallback: function isStatusCallback(req) {
+    const messageStatus = req.body.MessageStatus;
+
+    /**
+     * Twilio has received confirmation of message delivery from the upstream carrier,
+     * and, where available, the destination handset.
+     * @see https://www.twilio.com/docs/api/messaging/message#message-status-values
+     */
+    return messageStatus === 'delivered';
+  },
 };
--- a/lib/middleware/authenticate.js
+++ b/lib/middleware/authenticate.js
@@ -5,24 +5,40 @@ const logger = require('heroku-logger');
 
 const helpers = require('../helpers');
 
+const unauthorizedErrorMessage = 'Invalid or missing auth parameters. Unauthorized.';
+
 // TODO move to config file
-function validateAuth(user) {
+function validateBasicAuth(user) {
   return (user.name === process.env.DS_GAMBIT_CONVERSATIONS_API_BASIC_AUTH_NAME &&
     user.pass === process.env.DS_GAMBIT_CONVERSATIONS_API_BASIC_AUTH_PASS);
 }
 
+// TODO: We don't need this when Blink becomes the proxy between Twilio statuCallbacks and
+// Conversations. For now it's here for testing purposes, as the Twilio Messaging API doesn't
+// support setting Basic Auth in the statusCallback query param when POSTing a new message.
+function validateAPIKey(apiKey) {
+  return (apiKey === process.env.DS_GAMBIT_CONVERSATIONS_API_KEY);
+}
+
+function getAPIKey(req) {
+  return req.query.apiKey || req.body.apiKey;
+}
+
 function sendUnauthorizedResponse(req, res) {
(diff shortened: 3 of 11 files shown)


rapala61:
cc @sergii-tkachenko @aaronschachter. Would like to talk about this in our Dev Jam today.

lib/helpers/broadcast.js
```diff
@@ -0,0 +1,20 @@
...
+function getBroadcastId(req) {
+  return req.body.broadcastId || req.query.broadcastId;
+}
+
+/**
+ * Broadcast helper
+ */
+module.exports = {
+  /**
+   * parseBody - parses properties out of the body of a Facebook request
```
aaronschachter: Facebook?

lib/helpers/request.js
```diff
@@ -22,4 +22,14 @@ module.exports = {
     // TODO: Should be in a config constant
     return platform === 'customerio';
   },
+  isStatusCallback: function isStatusCallback(req) {
```
aaronschachter: Would `/lib/helpers/twilio` be a better home for this function? It's specific to Twilio. Or maybe just renaming as `isTwilioStatusCallback` ?

lib/middleware/authenticate.js
```diff
@@ -5,24 +5,40 @@ const logger = require('heroku-logger');
...
+const unauthorizedErrorMessage = 'Invalid or missing auth parameters. Unauthorized.';
+
 // TODO move to config file
-function validateAuth(user) {
+function validateBasicAuth(user) {
   return (user.name === process.env.DS_GAMBIT_CONVERSATIONS_API_BASIC_AUTH_NAME &&
     user.pass === process.env.DS_GAMBIT_CONVERSATIONS_API_BASIC_AUTH_PASS);
 }
 
+// TODO: We don't need this when Blink becomes the proxy between Twilio statuCallbacks and
```
aaronschachter: Spelling police 🚓 : `statuCallbacks`

lib/middleware/authenticate.js
```diff
@@ -5,24 +5,40 @@ const logger = require('heroku-logger');
...
+// Conversations. For now it's here for testing purposes, as the Twilio Messaging API doesn't
+// support setting Basic Auth in the statusCallback query param when POSTing a new message.
+function validateAPIKey(apiKey) {
```
aaronschachter: Capitalization police - why `API` vs `Api`?

lib/middleware/import-message/params.js
```diff
@@ -8,11 +8,21 @@ module.exports = function params() {
   return (req, res, next) => {
     logger.debug('POST /import-message', { body: req.body });
 
-    if (!helpers.request.isCustomerIo(req)) {
-      const error = new UnprocessibleEntityError('Invalid platform.');
+    // Customer.io's webhook import
+    if (helpers.request.isCustomerIo(req)) {
```
aaronschachter: Don't need to remove this now, but just checking: In theory, once we're committed to using the Twilio statusCallback, we can remove all Customer IO related code in Conversations?

lib/middleware/authenticate.js
rapala61: Are you asking why I named the function `validateAPIKey ` vs `validateApiKey`?

aaronschachter: Ya

lib/middleware/import-message/params.js
rapala61: Yes, but for now, since we still have the concept of the "backup plan" I decided to keep it.

lib/middleware/authenticate.js
rapala61: I don't feel strongly about either. I think is consistent with how I capitalize the word API in general. I'll change it to be more consistent with how we treat the IO acronym in `customerIo` within the request helper.

rapala61:
updated! @aaronschachter.

sergii-tkachenko:
@rapala61 
> URI can contain an API Key as a query parameter and be authenticated. Basic Auth in the statusCallback doesn't seem to be supported.

Hm. Could you double check that? Blink now would reject all requests that have no or incorrect HTTP auth credentials. Overriding this behavior would take time and testing.

rapala61:
@sergii-tkachenko As I mentioned over in Slack, turns out we can use Basic Auth. However, Twilio doesn't automatically send the `statusCallback` requests with the appropriate Authorization headers. For each request (message status change) we should expect this flow:

- Sends POST request to `statusCallback` URI, without `Authorization` headers.
- The request fails authentication (401) so we respond with a `WWW-Authenticate: Basic` header.
- Twilio retries the request including the `Authorization` header.

cc @aaronschachter

PR #144, opened 2017-09-27
Northstar support for Slack + Consolebot

### LINKED ISSUE #138: Create User middleware
Tasks
* [x] Create Northstar User if not found for the platformUserId
* [x] Modify Consolebot to allow creating a new User

aaronschachter:
* Slack support: Gets Northstar User by the email associated with Slack ID passed.
   * Doesn't create a new user if the Slack user's email doesn't exist in NS, to KISS and less code (we don't need to support creating users by email in production)
    * Adds a `SLACK_BOT_USER_OAUTH_ACCESS_TOKEN` - needed to query the Slack API to read a User's profile. Also needed to change the permissions in the Slack app to allow reading User profile (screenshot below)

* Consolebot: Pass a mobile number. Creates a new user if phone number doesn't exist (we do need to support this feature in production, hence adding to Consolebot)
    * Completes last task in #138, resolves #138

## Testing
This goes hand in hand with https://github.com/DoSomething/gambit/pull/965, unfortunately they both kind of need to be running at the same time. It felt a bit too complicated to support passing `phone` on the Gambit Campaigns side as well as `userId` when eventually we'd just rip out the `phone` param once this feature in Conversations was complete.

I've set the `SLACK_BOT_USER_OAUTH_ACCESS_TOKEN` and `SLACK_OAUTH_ACCESS_TOKEN` in Staging, and will deploy this branch as well as https://github.com/DoSomething/gambit/pull/965 to Campaigns Staging to test.

To test locally, you need to run both apps in these branches, but then also run Gambit Adapters if you'd like to test Slack with @gambit-dev. Or, consolebot would work if you run this branch and also run Campaigns https://github.com/DoSomething/gambit/pull/965 locally.
 
## Screenshots
screen shot 2017-09-27 at 9 42 24 am


aaronschachter:
This branch and https://github.com/DoSomething/gambit/pull/965 have been deployed to staging, can see the Northstar ID in logs when chatting with @gambit-staging over Slack.

rapala61:
A question about the current approach. Would it make it more convenient to move the fetching of the Slack user's email to Adapters? This way we could add it as a property in the POST to conversations. I'm aware this is not a crucial item, but just curious if it would be a low level effort to streamline it that way, instead of adding two external calls to Conversations (fetchSlackUserBySlackId, fetchUserByEmail).

aaronschachter:
I did consider it, especially since [Adapters uses the Slack RTM client](https://github.com/DoSomething/gambit-adapters/blob/4b786440d6ae3440e0470ac90879dbdc0a271647/app/controllers/slack.js#L23), which has a handy [`getUserById`](http://slackapi.github.io/node-slack-sdk/reference/SlackDataStore#SlackDataStore+getUserById). Here's the rationale for why I didn't go that route, and definitely up for discussion / feedback:

* Ideally we should be querying Northstar by Slack ID, not by the email associated with the Slack account. Slack ID does exists as a [User property in Northstar](https://github.com/DoSomething/northstar/blob/be3d3d9b9634001704bc9c6a79e8a4851faa642a/app/Models/User.php#L63) -- there just isn't support to query by it yet (cc @DFurnes). Once there is, we should be doing that.
    * If we passed `email` as a param to `POST /receive-message` -- it'd make more sense for it to be stored as our `Conversation.platformUserId`, since the Slack User ID (`slackId`) we pass would no longer be used for anything. Buuut if a Slack user changes their email: I'd argue it should not change their Northstar ID. (e.g. If I'm connected to Northstar through Facebook and change my Facebook email, my Northstar account shouldn't change). If we ever make the Gambit Slack app public for external teams to use, the Slack ID should be the source of truth, not the email associated with the Slack ID. 

* I think it makes sense to deprecate Adapters completely at this point; The custom functionality we use for Slack for keywords list (and viewing the rendered templates as per Campaign) would be better suited as a web view in Gambit Admin. e.g. Clicking "view messages" button in one of the attachments of a "keywords" Gambit Slack response returns something like 20 different templates. I think a web app would make it easier to see which campaigns and keywords are available, as well as drilling down into all the available templates.
     * To deprecate Adapters for Slack, we'd need to either add the RTM client to Conversations, or [use incoming webhooks](https://api.slack.com/apps/A6P00Q01J/incoming-webhooks) instead of the RTM client. Using incoming webhooks would be a consistent implementation to how we're using Twilio (and one day in v1.2, Messenger)

rapala61:
Got it! Thanks for the thorough comment. I didn't know we are tracking 3rd party platform ids in Northstar. It makes sense now.

Issue #37, opened 2017-07-25
Campaign caching


aaronschachter:
On app start, the Conversations API queries the Gambit Campaigns `/campaigns` endpoint to update or create documents in our Mongo `campaigns` collection. The Receive Message route queries the `campaigns` collection to do things like select a Campaign upon a MENU keyword, or prompt whether the User wants to continue the Campaign in their Conversation.

There's a bit of cleanup / logic we need to add:

* [x]  Update the status of Campaigns that are no longer running on the Gambit. The Gambit `/campaigns` endpoint only returns active Campaigns, so if a match for a document isn't found in the GET query, it needs to be updated as closed (and unavailable for Conversation)

* [x] We need a way to refresh the Campaigns cache without restarting app -- regularly polling Campaigns would be ideal, but it'd also be useful to force a cache clean for a Campaign

* [x] When we receive a Signup message for a Campaign, we post it to the Gambit Campaigns `/receive-message` endpoint (https://github.com/DoSomething/gambit/issues/939). Gambit Campaigns is querying out to Phoenix and Contentful when rendering a reply message, but we shouldn't have to do that if we have the cached rendered Campaign messages already in the Conversations API. Tasks to improve
   * [x] Store all campaign templates in Conversations model - https://github.com/DoSomething/gambit-conversations/pull/146
   * [x] Return the updated Signup (with its loaded Draft Submission) in Gambit Campaigns `POST /receive-message` - https://github.com/DoSomething/gambit/pull/967
   * [x]  When Conversations receives the Campaigns  `POST /receive-message` response, return the cached template for the response.replyTemplate, replacing any `{{quantity}}` tags with the Signup values returned
   * [x] On the Gambit Campaigns side, move the `sendResponseForSignup` calls outside of the `helpers.renderMessageTemplate` promise to avoid querying Phoenix/Contentful in every request

aaronschachter:
Added regular polling in #125. As a worst case, we could restart the app if we needed to force refresh immediately.

PR #146, opened 2017-09-28
Gambit Campaigns API changes

aaronschachter:
* Changes per `POST /receive-message` response in https://github.com/DoSomething/gambit/pull/967

* Stores all rendered Campaign templates to prep for final task in #37


aaronschachter:
This branch and https://github.com/DoSomething/gambit/pull/967 have each been deployed to staging, verified that Conversations + Campaigns integration is working as expected.

app/models/Campaign.js
```diff
@@ -17,14 +17,26 @@ const campaignSchema = new mongoose.Schema({
   keywords: [String],
   topic: String,
   templates: {
```
rapala61: Not a priority, but it would be nice to also port the `errorOccurred` template field.

aaronschachter: Great call! I forgot about that one. In theory, we shouldn't really need that anymore -- we only used it when an emoji would doom a Reportback Submission to fail, which was pre-Rogue. I have a feeling we can get away not bothering with the Blink Suppress Retry headers for now (we're currently not sending them -- although we could open an issue to do so)

Issue #149, opened 2017-09-29
Add FAQ / General Rivescript


aaronschachter:
https://docs.google.com/document/d/1CJQBseG8xBPaYbrGgfApZK_YCEvl-OjJy2X9UnyZKPI/edit?disco=AAAABREL0Yo&ts=59e646a9

aaronschachter:
Discussed this with @justkika -- we're going to move our Question and Crisis triggers into Contentful, so I'll be opening up a PR to remove them from the codebase.

Things we do want to keep in the codebase:
* Substitutions defined in the [`/brain/index`](https://github.com/DoSomething/gambit-conversations/blob/d5853b3df67f496adc50b8152a4b54f93a6bcf12/brain/index.rive)
* Yes/No triggers for Campaign Signup [`brain/campaign`](https://github.com/DoSomething/gambit-conversations/blob/d5853b3df67f496adc50b8152a4b54f93a6bcf12/brain/campaign.rive)

We should also add a Rivescript field to our Campaign Contentful field type to support FAQ per Campaign.

aaronschachter:
This one feels ready to close because we've got the core Rivescript code defined in the `app/brain`. Additional triggers are managed within Contentful.

PR #88, opened 2017-08-24
Ride Or Die aka The Seatbelt Game

aaronschachter:
Resolves #80: Adds our SMS Game script as a Rivescript file.

* Each level in the game is implemented as a [Rivescript topic](https://www.rivescript.com/docs/tutorial#topics) to support level-specific triggers (some levels require the User to send a `Y` to advance, other levels require sending `'A'` or `'B'`)

Issue #80, opened 2017-08-22
Ride or Die Rivescript


aaronschachter:
As a developer, I want to support [Ride or Die](https://docs.google.com/a/dosomething.org/document/d/1KAIKXGeiyeHGqdYxqm0JQQZfFj7ZA6M1X8IFGK5pX68/edit?usp=sharing) conversations via Rivescript topics.

PR #150, opened 2017-09-29
Load Rivescript from Contentful

aaronschachter:
Queries Contentful for entries of a new content type, Rivescript, and loads it on app start.

* Removes all non-macro triggers out of our codebase, to be manually entered in Contentful (I've started on this for #149, also moved Ride and Seek to avoid hardcoding in #80, #88) 

* Adds a `match` property to a Message to indicate the Rivescript trigger that matched in the inbound message text:

screen shot 2017-09-29 at 3 34 04 pm

## TODO
* [ ] Extract fetching/writing the Rivescript files out of `server.js` into a separate function
* [ ] Handle when a Rivescript entry gets deleted (for now, let's just not delete them)
* [ ] Add a Contentful webhook to post updates to the bot's rivescript  -- nice to have, at a minimum we could restart the app (which we'd need to do anyway to deploy any copy changes in code)


server.js
```diff
@@ -23,6 +23,34 @@ const Campaign = require('./app/models/Campaign');
...
+
+// Fetch rivescript files from Contentful.
+contentful.fetchRivescripts()
+  .then((entries) => {
+    entries.forEach((entry) => {
+      const id = entry.sys.id;
+      const script = entry.fields.rivescript;
+      const filename = `${dir}/${id}.rive`;
+      // Write them.
+      fs.writeFile(filename, script, ((err) => {
```
rapala61: I wonder if a combination of [mock-fs](https://github.com/tschaub/mock-fs) and/or some lightweight in-memory DB could work here without relying on Heroku's file system. Just a thought!

aaronschachter: Good ideas! Worth giving a look when it's time to extract this into a different function in the upcoming QA / cleanup sprints.

Issue #116, opened 2017-09-11
Query Broadcasts by entry ID, not fields.broadcastId


aaronschachter:
We may purpose the broadcastId for readability (or create a field like `name` or `title`), but the Broadcast ID we're adding to Customer.io broadcasts should correspond with the Broadcast entry ID, not the `broadcastId` field.

Refs 
* https://github.com/DoSomething/gambit-conversations/pull/48#pullrequestreview-61336107
* https://github.com/DoSomething/gambit-conversations/pull/106

PR #151, opened 2017-09-29
Adds `broadcast-settings` route.

### LINKED ISSUE #142: Add route to generate the Webhook body, given a `broadcastId`.
# Spec
- End point will be protected by Basic Auth.
- `broadcastId`is required.
    - If `broadcastId` doesn't exist, return an error.

# Tasks
- [x] Remove all Customer.io code used to support the `Backup solution` in #139. - PR #143.
- [x] Create `broadcast-settings` endpoint that generates a JSON response with all necessary settings to execute a new broadcast. - PR #151.
- [x] Update endpoint to documentation. - PR #143.

rapala61:
## Summary
This is a medium size PR.

It adds a new protected route to get the Customer.io Webhook setup settings for the given `broadcastId`.

## How to test
GET settings for the broadcastId: `tacosfest`.

```
curl -X "GET" "http://localhost:5100/api/v1/broadcast-settings/tacosfest" \
     -H "Authorization: Basic cHVwcGV0OnRvdGFsbHlzZWNyZXQ="
```
> You might need to change the Authorization header's value.

## Relevant Issues
Fixes #142
Adds more documentation for #123 

## Tasks
- [x] Update LastPass `Gambit-conversations-env-keys` secured note with new env variable values.
- [x] Update Heroku with env variables.

### DIFF
--- a/documentation/endpoints/broadcast-settings.md
+++ b/documentation/endpoints/broadcast-settings.md
@@ -1,0 +1,56 @@
+# Broadcast Settings
+
+```
+GET /api/v1/broadcast-settings/<broadcastId>
+```
+Returns the settings necessary to setup a Customer.io Triggered Campaign **after** the broadcast is created in Contentful.
+
+## Authentication
+
+The endpoint expects you to use [Basic Auth](../authentication.md). Example url using the browser: `http://protectedName:protectedPass@localhost:5100/api/v1/broadcast-settings/100`
+
+## Examples
+
+
+<details>
+<summary><strong>Example Request</strong></summary>
+
+GET settings for the broadcastId: `tacosfest`.
+
+```
+curl -X "GET" "http://localhost:5100/api/v1/broadcast-settings/tacosfest" \
+     -H "Authorization: Basic cHVwcGV0OnRvdGFsbHlzZWNyZXQ="
+```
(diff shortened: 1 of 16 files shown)


documentation/endpoints/broadcast-settings.md
```diff
@@ -0,0 +1,56 @@
...
+```
+</details>
+
+<details>
+<summary><strong>Example Response</strong></summary>
+
+```
+{
+  "broadcast": {
+    "broadcastId": "tacosfest",
```
aaronschachter: I think our general standard practice at DS is generally to use `id` when it's a nested object (e.g. https://www.dosomething.org/api/v1/signups).. which is how I formatted the [G-Campaigns Recieve Message response](https://github.com/DoSomething/gambit/blob/develop/documentation/endpoints/receive-message.md).
I get that this is the `broadcastId` field and not the entry id, but maybe it makes sense to take on https://github.com/DoSomething/gambit-conversations/issues/116 next so this can be the legit `id` (vs `broadcastId` field)?

rapala61: Yeah, I decided to keep the contentful querying consistent for now, but I'm aware of the changes in #116.

PR #152, opened 2017-10-02
encode characters '!()* in statusCallback URI

rapala61:
## Summary
Fixes statusCallback URI triggering [invalid-statuscallback-url 21609](https://www.twilio.com/docs/api/errors/21609#invalid-statuscallback-url) Twilio errors. Implements a more stringent form of `encodeURIComponent` suggested in [MDN Docs for encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent).


rapala61:
Encoding doesn't fix the problem. We probably will have to change password.

PR #157, opened 2017-10-06
Fixes Relay of Blink suppress headers and unrejected error in Conversations.

### LINKED ISSUE #156: Receiving legacyError from Campaigns breaks Conversations. Relay Blink suppress headers.
## BUG

When Campaigns responds with a legacy error to Conversations. We should relay the error back to Blink, together with the corresponding Blink suppress headers so that Blink doesn't keep retrying a known bug in the system.

As of now, receiving a legacy error breaks Conversations due to the error not being properly rejected, resulting in the error:

```
2017-10-03T22:07:04.453006+00:00 app[web.1]: code=500 response#message="Cannot read property 'signup' of undefined" level=debug message=sendResponseWithStatusCode
```

## How should it work?
Conversations should relay the Blink suppress headers from the Campaigns response to Blink. It should also properly reject this error to prevent the `Cannot read property 'signup' of undefined` error.

rapala61:
## BUG
Rejects the error when receiving an error from Campaigns. Also relays the Blink suppress headers from Campaigns to Blink to prevent unwanted retries.

## How should it be tested
- 👀 

## Related issues
Fixes #156

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -90,6 +90,11 @@ module.exports.sendErrorResponse = function (res, err) {
   if (!message) {
     message = err;
   }
+
+  if (err.response && err.response.get(config.blinkSupressHeaders)) {
+    exports.addBlinkSuppressHeaders(res);
+  }
+
   return this.sendResponseWithStatusCode(res, status, message);
 };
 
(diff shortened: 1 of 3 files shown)


lib/helpers.js
```diff
@@ -91,6 +91,11 @@ module.exports.sendErrorResponse = function (res, err) {
   if (!message) {
     message = err;
   }
+
+  if (err.response && err.response.get(config.blinkSupressHeaders)) {
```
sergii-tkachenko: Sorry, I'm not sure I'm following the logic for `err.response.get(config.blinkSupressHeaders)`.
This retransmits the suppress header from Gambit Campaigns through Gambit Conversations to Blink, right?

rapala61: Correct!

sergii-tkachenko: Thanks, a great catch, Rafa! No, really.
One thing I'd recommend though is to add the comment explaining what's going on here. It took me some mental capacity understand this code's purpose.

sergii-tkachenko: Also, I guess I should start reading PR descriptions first :neckbeard:.

PR #166, opened 2017-10-13
Send suppress Blink headers when user has no valid phone number.

### LINKED ISSUE #162: Temporary feature request: filter out members with no phone number on /send-message
### Existing behavior
When Blink posts a `signup_post` that belongs to a user with no phone number to Gambit's `/send-message`, the request fails with:
```
response_status=500 response_status_text="Internal Server Error" {"message":"The string supplied did not seem to be a phone number"}
```
As a side-effect, Blink will enter this request into undesirable retry loop.

### Temporary solution
Either filter out (ignore) this kind of messages or suppress undesired retries in Blink by setting `x-blink-retry-suppress: true` header on the response.

### Optimal solution
When Blink has a feature that would allow extending known messages with data from other resources (in this case extending Rogue's `signup_post` with Northstar's profile user info), it will be able to filter out signup posts of users with no phone number on their profile and not send these requests to  Gambit's `/send-message`.
Related feature request in Blink: **pending**.

### Related tickets
This came up during https://github.com/DoSomething/blink/pull/147

rapala61:
# What does this PR do?
It adds the `x-blink-retry-suppress` headers to the error response to Blink when the user's Northstar profile:
- Is not found.
- Lacks a valid phone number.

# How to review
- 👀 
- send a message using an invalid Northstar Id. The error response should include the suppress headers.
- send a message using a valid Northstar id of an user with no phone number. The error response should include the suppress headers.

# Relevant tickets
Fixes #162

### DIFF
--- a/lib/middleware/send-message/user-get.js
+++ b/lib/middleware/send-message/user-get.js
@@ -20,6 +20,7 @@ module.exports = function getNorthstarUser() {
            */
           req.platformUserId = helpers.formatMobileNumber(user.mobile);
         } catch (error) {
+          helpers.addBlinkSuppressHeaders(res);
           logger.error('getNorthstarUser: Fatal error formatting Northstar user\'s mobile.');
           return helpers.sendErrorResponse(res, error);
         }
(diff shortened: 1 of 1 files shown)


aaronschachter:
Code looks good, but can we confirm this fixes #162 per question in https://github.com/dosomething/gambit-conversations/issues/162#issuecomment-335921486 ? Will Blink post to Conversations `/send-message` when a Signup is created via keyword or broadcast in Conversations `/receive-message`? cc @sergii-tkachenko

sergii-tkachenko:
Will test it on staging now.

sergii-tkachenko:
👍

PR #168, opened 2017-10-16
Add test credentials support for our Twilio library

rapala61:
# What does this PR do?
It adds support to switch our Twilio library to use test credentials instead of production.

# How to test? 
Will test in Staging. Conversations responses should not arrive to your phone.


rapala61:
Tested in Staging and the outbound message did not arrive to my device + no errors in logs. 👍

config/lib/twilio.js
```diff
@@ -1,8 +1,20 @@
 'use strict';
 
+const useTwilioTestCreds = process.env.DS_GAMBIT_CONVERSATIONS_SHOULD_USE_TWILIO_TEST_CREDS || 'false';
```
aaronschachter: A bit nitpicky, but what do you think about just naming this var `DS_GAMBIT_CONVERSATIONS_USE_TWILIO_TEST_CREDS`? or DS_GAMBIT_CONVERSATIONS_IS_TWILIO_TEST`? It seems a little wordier than it needs to be.

lib/twilio.js
```diff
@@ -10,7 +10,8 @@ module.exports.postMessage = function (phone, messageText) {
     body: messageText,
   };
 
-  if (config.messagingServiceSid) {
+  // Twilio test credentials don't support messagingServiceSid.
+  if (config.messagingServiceSid && config.useTwilioTestCreds !== 'true') {
     payload.messagingServiceSid = config.messagingServiceSid;
   } else {
     payload.from = config.fromNumber;
```
aaronschachter: We only use a From number for these tests now, since we're no longer going to support using our Mobile Commons Twilio account.
Would it be cleaner to just check whether `config.useTwilioTestCreds` is true and we have a from? If that condition isn't met, return an error that we can't run the test.

rapala61:
Thank for the feedback @aaronschachter, it's always welcome. I'm going to have to close this PR and open a new one with the changes you suggested. Somehow Wercker is refusing to rebuild this PR on updates.

PR #175, opened 2017-10-18
Implements contentful caching when importing messages

rapala61:
# What does this PR do?
It caches the result of the request to Contentful when fetching a broadcast for importing messages. The cached broadcast is kept in memory for 300s (5 min) by default. With our planned throughput of ~ 3K requests a minute when importing messages, we should be saving about ~15K requests to Contentful every 5 min!

# How to test?
- change the `DS_GAMBIT_CONVERSATIONS_BROADCASTS_CACHE_TTL` to 20.
- import a message through the `import-message` route.
    - You should see an info log that outputs `Broadcasts cache miss`.
    - A second request immediately after should result in a log that outputs `Broadcasts cache hit`.
    - After the 20 seconds, a third request should yield `Broadcasts cache miss`.


lib/middleware/import-message/broadcast.js
```diff
@@ -47,11 +69,20 @@ module.exports = function getBroadcast() {
       return helpers.sendErrorResponse(res, error);
     }
 
-    asyncProperties.push(parsePropertiesFromBroadcast(req, res));
+    return cache.get(req.broadcastId)
+      .then((broadcast) => {
+        if (broadcast) {
+          logger.info(`Broadcasts cache hit. Broadcast id: ${req.broadcastId}`);
```
aaronschachter: Let's use heroku-logger for what it should be best for, creating different fields in Logstash parsing by object properties. Also maybe `debug` would be better for less noise on production?

e.g. `logger.debug('broadcast.cache hit', { broadcastId: req.broadcastId })`

Issue #121, opened 2017-09-13
Global keywords: sending messages when conversation is paused


aaronschachter:
Refs https://support.twilio.com/hc/en-us/articles/223182208-Industry-standards-for-U-S-short-code-HELP-and-STOP - our STOP, HELP keywords need to return messages even a Conversation has been paused.

* [x] HELP, INFO
* [x] STOP, END, QUIT, CANCEL, UNSUBSCRIBE 
* [x] Crisis messages

aaronschachter:
Our HELP trigger needs to be implemented in a way that we still send the reply message when a User's conversation is paused.

aaronschachter:
We have the same issue with the STOP keyword: it needs to still return the unsubscribed message if the Conversation is paused.

PR #178, opened 2017-10-20
Global keyword support

aaronschachter:
Refs #121: Adds HELP and INFO as keywords that return an industry standard info message, regardless of the Rivescript topic a Conversation is set to.

Refactors Receive Message middleware to determine whether a Conversation is paused and when to send a message back (e.g. when User sends HELP).  This was done before via a separate Rivescript topic called Support to filter the various triggers were define in a `random` topic in another RS file, which I'll be deprecating in a future PR.

* This means the NEVERMIND trigger won't exit you out of Support anymore. The trick to exit support is to send a Campaign Keyword. 
* Another future PR coming to refactor the QUESTION keyword to return the User's current Campaign's `supportMessage` template, instead of the Rivescript reply we have defined `brain/support`.

How to test:
* Confirm that sending INFO or HELP sends the Info template back for both paused and unpaused Conversations. Will deploy to staging for testing via SMS.


app/routes/receive-message.js
```diff
@@ -36,13 +37,17 @@ router.use(createUserIfNotFoundMiddleware());
 router.use(loadInboundMessageMiddleware());
 router.use(createInboundMessageMiddleware());
 
-// If Campaign keyword was sent, update Conversation campaign and send continueCampaign reply.
+// Updates Last Messaged At, Subscription Status, Paused.
+router.use(updateUserMiddleware());
```
aaronschachter: Moving this before `rivescriptMiddleware` introduces a bug where STOP isn't registering, because we're relying on the Rivescript `subscriptionStatusStop` macro to be returned. Will be pushing up a fix to this branch shortly.

brain/random.rive
```diff
@@ -16,19 +19,4 @@
...
-- Text back your question and I'll try to get back to you within 24 hrs.\n\nIf you want to cancel\s
-^ your request without waiting for a response, reply with "nevermind".{topic=support}
-
-+ q
-@ question
-
-+ help me [*]
-@ question
-
-+ [*] you help me [*]
```
aaronschachter: We can move and fine-tune these variations `[*] you help me [*]`, `help me [*]`  in Contentful.

brain/random.rive
```diff
@@ -16,19 +19,4 @@
 + [*] i am [feeling] (depressed|bipolar|freaking out) [*]
 @ crisis
 
-// General:
-
-+ question
-- Text back your question and I'll try to get back to you within 24 hrs.\n\nIf you want to cancel\s
-^ your request without waiting for a response, reply with "nevermind".{topic=support}
```
aaronschachter: Setting the User's topic to Support here via Rivescript to pause and unpause was causing the order of operations issue described in https://github.com/DoSomething/gambit-conversations/pull/178#discussion_r146044590

726c9c4ff147bfa52d2815e7e9a0e1a0d57429d4 refactors pausing Conversation to be handled via macro with middleware.

aaronschachter:
Updated and ready for review, will be deploying to staging to test.

* Refactors how we update Conversation topic to `'support'`: Instead of changing the topic in a Rivescript reply, we add a new `supportRequested` macro to handle in middleware 
* Adds `support-requested` middleware to check for the `supportRequested` macro, renames `support` middleware as  `support-message`
* Splits out the `rivescript` middleware into separate files:
    * `rivescript-reply-get`: Posts inbound message to Rivescript bot and saves the response to `req`
    * `template-rivescript`: Sends the text returned from Rivescript as outbound reply message

rapala61:
@aaronschachter 2 questions. Should I be seeing my aurora profile update when I send STOP or LESS? Also, should I be getting the same copy for both HELP and INFO?

aaronschachter:
1 - Yes, SMS Status and Paused update accordingly in Aurora. 
screen shot 2017-10-20 at 2 02 47 pm

2 - Yep! https://support.twilio.com/hc/en-us/articles/223182208-Industry-standards-for-U-S-short-code-HELP-and-STOP

rapala61:
The test passed 🌟 💯 .

2 things I noticed, they might be working as spec, but just noting them:

- sending Less will keep me in sms_status=less unless I send STOP and then another message, only then I become active again.
- I stay in paused mode even after someone sends me a text from Front, it gets unpaused when I text back something.

aaronschachter:
> sending Less will keep me in sms_status=less unless I send STOP and then another message, only then I become active again.

As expected - that logic happens in the [ user-update` middleware.](https://github.com/DoSomething/gambit-conversations/blob/1f6c9f1cc8f21f3195e752057bc0fdbcad4fcdf1/lib/middleware/receive-message/user-update.js#L25).  Currently in MC we have a WEEK keyword that would transition someone from `less` to `active`, but don't have support for that yet as is. We'd add it as a new keyword/macro to our `brain/keywords.rive`.

> I stay in paused mode even after someone sends me a text from Front, it gets unpaused when I text back something.
Do you mean your Aurora profile? That would check out -- we only update a Northstar User when we receive a message (not updating Northstar User in the `POST /send-message` request that delivers the Front message to our User). It'd make sense to though! I remember seeing chat about creating a separate endpoint for Front `send-message` vs the Blink `send-message`.  Sounds like a good post TGM refactor.

lib/middleware/receive-message/rivescript-reply-get.js
```diff
@@ -11,15 +11,8 @@ module.exports = function getRivescriptReply() {
...
-
-        return req.conversation.setTopic(rivescriptRes.topic);
-      })
-      .then(() => {
-        if (helpers.isMacro(req.rivescriptReplyText)) {
-          return next();
-        }
-
-        return helpers.rivescriptReply(req, res, req.rivescriptReplyText);
+        req.rivescriptTopic = rivescriptRes.topic;
```
aaronschachter: 💣

lib/middleware/receive-message/template-rivescript.js
```diff
@@ -0,0 +1,15 @@
...
+
+const helpers = require('../../helpers');
+
+module.exports = function sendRivescriptReply() {
+  return (req, res, next) => {
+    if (helpers.isMacro(req.rivescriptReplyText)) {
+      return next();
+    }
+
+    return req.conversation.setTopic(req.rivescriptReplyTopic)
```
aaronschachter: 💣

Issue #171, opened 2017-10-17
Move Campaigns.sync and interval to worker process


rapala61:
# User Story
As an O'Dev dev, I want to move the Campaigns cache sync to a separate worker Dyno so that when it updates the cache, it doesn't run on all Dynos at once.

aaronschachter:
I've been thinking about this in light of the caching introduced in #175,and have an idea about an alternative approach:

* Step 1: The `Campaign.sync` queries the GamCam `GET /campaigns` index to create a cached keywords object, where each property is an published keyword, and its value is it's Campaign ID:
```
keywords: {
  bookbot: 2900,
  bumblebot: 2299,
  mirrorbot: 7,
  ..,
}
```
* Step 2: Our `campaign-current` and `campaign-keyword` middleware for Receive Message could maintain cached Campaign objects (querying the GamCam `GET /campaigns/:id` endpoint when necessary). `campaign-keyword` would be checking the keyword cache for the campaignId stored witha  keyword, whereas `campaign-current` checks the current Conversation.campaignId

* Step 3: Profit? 💰 

This would essentially deprecate the Mongoose `GET /campaigns` route, which is only used by Gambit Admin (which could query Gambit Campaigns API anyway, so that's not a biggie)

rapala61:
I don't see why we couldn't use the same approach. That being said, we have to take our sweet time refactoring out our dependence on the cached campaigns inside the mongo db. Our assumption that this object exists is in quite a few middleware: `campaign-keyword`, `campaign-current`, `campaign-closed`, `campaign-continue`, `campaign-menu`, etc.

rapala61:
This was fixed by @aaronschachter in #182 and #184.

PR #182, opened 2017-10-23
Campaign caching + topic cleanup

aaronschachter:
* Caches a Gambit Campaign `GET /campaign/:id` response for a Campaign to avoid regularly syncing every hour in #171. 
   * Refactors `campaign-current` middleware to get the current Campaign from cache or from Gambit Campaigns if cache expired, via `gambitCampaigns.getCampaignById`

 * Refactors Campaign model to match the Gambit Campaigns response (instead of string properties in `templates`, use objects with `rendered`, `raw`, and `override` properties)

* Removes Campaign-topic specific Rivescript. We aren't using this functionality currently, and can consider adding a Rivescript field to a Contentful Campaign if we need Campaign-specific triggers.

TODO:
The `campaign-menu` and `campaign-keyword` are still loading `req.campaign` from Mongoose, and rely on `Campaign.sync` at system start to get active Campaigns and Keywords. The next step to close out #171 would be to cache the Gambit Campaigns `GET /campaigns` index response to keep a cache of active Keywords (to be used by `campaign-keyword`) and active Campaigns (to be used by `campaign-menu`)


config/lib/gambit-campaigns.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+module.exports = {
+  cacheTtl: process.env.DS_GAMBIT_CONVERSATIONS_CAMPAIGNS_CACHE_TTL || 300,
```
rapala61: Can we add this bad boy in the `.env.example`?

aaronschachter: Good call! Updated.

lib/gambit-campaigns.js
```diff
@@ -66,3 +70,29 @@ module.exports.postReceiveMessage = function (req) {
...
+      .then(res => cache.set(`${campaignId}`, res))
+      .then(campaign => resolve(campaign))
+      .catch(error => reject(error));
+  });
+}
+
+module.exports.getCampaignById = function (campaignId) {
+  logger.debug('getCampaignById', { campaignId });
+
+  return cache.get(`${campaignId}`)
```
rapala61: I'm also curious, why use templating here vs just the variable? I faced the same debacle when implementing in broadcasts and went with just sending the variable.

aaronschachter: The Cacheman module was throwing an error that we needed to pass a string or an array.

rapala61: Ha, I guess it didn't give me the error since the id is a string.

Issue #55, opened 2017-08-10
Disable CORS


aaronschachter:
CORS was enabled in https://github.com/DoSomething/gambit-conversations/pull/23 to quickly develop the Gambit Admin prototype. Once (https://github.com/DoSomething/aurora/issues/157) is done, we should disable it.

PR #184, opened 2017-10-23
Active Campaign + Keyword cache

aaronschachter:
Follow up to #182: Maintains a cache of active Campaign id's as well as a map of their keywords, deprecating the Campaign model, resolving #171.

After verifying all endpoints work as expected, I will push to this branch up to staging for more testing.

TODO: Hide the Gambit Admin Campaign links for now (or hotfix Gambit Admin / Campaigns to allow Gambit Admin to query the Gambit Campaigns API (#55)


lib/gambit-campaigns.js
```diff
@@ -96,3 +181,53 @@ module.exports.getCampaignById = function (campaignId) {
     })
     .catch(err => err);
 };
+
+/**
+ * @param {string} keyword
+ * @return {Promise}
+ */
+module.exports.getCampaignByKeyword = function (keyword) {
```
rapala61: Why not make this method work like `getCampaignById `? I don't remember from the top of my head if the `Campaigns` endpoint supports getting a campaign by keyword, if not, adding that endpoint could be part of this task. I think we should keep favoring atomic caches vs batch caches like `activeCampaigns`.

lib/gambit-campaigns.js
```diff
@@ -4,36 +4,113 @@ const superagent = require('superagent');
...
+      .then(res => cache.set(activeCampaignsCacheKey, parseCampaignsIndexResponse(res)))
+      .then(activeCampaigns => resolve(activeCampaigns))
+      .catch(error => reject(error));
+  });
+}
+
+/**
+ * @return {Promise}
+ */
+function getActiveCampaigns() {
```
rapala61: This operation is still very expensive to be ran on each Dyno after the TTL is expired. My suggestion would be to add support in `Campaigns` to return an array of active campaigns like `[{ id: 1234, keyword: 'etc' }]`. We can cache this array as `activeCampaigns`. This way, even if it runs on all Dynos when it expires, it would be less expensive to perform. This also favors the atomic caches we are already making per campaign which is more consistent and easier to refactor when we introduce a cache layer like Redis in the future.

lib/middleware/send-message/campaign.js
```diff
@@ -25,7 +25,7 @@ module.exports = function sendCampaignTemplate() {
           return helpers.sendErrorResponse(res, err);
         }
 
-        req.outboundMessageText = campaign.templates[req.outboundTemplate];
+        req.outboundMessageText = campaign.templates[req.outboundTemplate].rendered;
```
rapala61: Nice catch.

lib/gambit-campaigns.js
aaronschachter: > My suggestion would to add support in` Campaigns ` to return an array of active campaigns

By `Campaigns` -- do you mean the Gambit Campaigns API? The Gambit Campaigns [GET /campaigns](http://ds-mdata-responder-staging.herokuapp.com/v1/campaigns) response does return id, keywords, and status. I'm working on a branch in that repo as well to make a single Phoenix APi request instead of one request per Campaign [here](https://github.com/DoSomething/gambit/compare/campaigns-index?expand=1), so it's not such a burden on the Gambit Campaigns side..

lib/gambit-campaigns.js
aaronschachter: Gambit Campaigns API doesn't have a `keyword` endpoint, but Contentful does. When we'd get a `campaignId` field back from Contentful, we'd still need to know whether that Campaign is active.

The idea behind having a cache of active keywords is that we'd be cutting down on the number of requests we make to see whether a given string is a keyword. With the approach you're suggesting, if someone texts in words like "Sweet" "K" "Whatev" -- we'd keep a `keyword` cache of each word with a null `campaignId`?

lib/middleware/send-message/campaign.js
aaronschachter: 💣 from #182

lib/gambit-campaigns.js
rapala61: Not necessarily, we could make the caching mechanism a little smarter and just cache those responses that do match a campaign, otherwise we wouldn’t. A get request to find that out if a text is a valid keyword is not really expensive in the scope of things, if it’s actually cached (this is a valid keyword), we wouldn’t even make the get request. 

Possible implementation: That single get request based on the keyword could return a 200 (campaign), 422 (not a valid keyword), a 404? (Campaign closed).

lib/gambit-campaigns.js
rapala61: Yeah, I meant Gambit Campaigns. My suggestion there was in case we wanted to go the way of keeping a cache of all keywords and their respective keywords and ids. This way we always know is a text we receive from an user is indeed a `keyword`. Then we would use the `id` value to `getCampaignById` and set an atomic cache on that campaign we get back from Gambit Campaigns.

lib/gambit-campaigns.js
rapala61: Alternatively, in light of [my comment above](https://github.com/DoSomething/gambit-conversations/pull/184#discussion_r146700226). We could make the `getActiveCampaigns` function a lot leaner by just getting back a simple array with objects that contain the campaign id and keywords. This of course needs the work you mentioned you are doing in Gambit Campaigns API.

lib/gambit-campaigns.js
aaronschachter: The `getCampaignByKeyword` function does work this way -- it calls the `getCampaignById` if a keyword exists in the `activeCampaigns.keywords` object to return the atomic cache.

I've just pushed up changes to `parseCampaignsIndexResponse` to store `campaignIds` as array of ID's of the Campaigns that have a non-`closed` status, instead of a property  `activeCampaigns` that stores the response.  As far as being an expensive operation to run regularly: I think the overhaul of https://github.com/DoSomething/gambit/pull/986 should put us in a better spot?

lib/gambit-campaigns.js
aaronschachter: Updated to do this ☝️

lib/middleware/receive-message/campaign-menu.js
```diff
@@ -11,7 +11,7 @@ module.exports = function campaignMenu() {
 
     // Find a random Campaign to prompt for Signup.
     // Eventually query Signups to find Campaigns that are new to User, within their interests, etc.
-    return Campaign.findRandomCampaignNotEqualTo(req.conversation.campaignId)
+    return gambitCampaigns.getRandomActiveCampaignNotEqualTo(req.conversation.campaignId)
```
rapala61: Just a heads up. I think there is a small edge case here. It would happen when the conversation has just been created and `req.conversation.campaignId` would be `undefined`.

aaronschachter:
Good call on both! Going to merge and make those changes in a new branch.

PR #207, opened 2017-11-01
Set sms_status instead of email

### LINKED ISSUE #202: New User tweaks
If creating a new user when we receive a message:

* [x] Don't send a [`mobile.import` email](https://github.com/DoSomething/gambit-conversations/blob/0.8.1/lib/northstar.js#L69) (no longer required)
* [x] Set their SMS status to `active`.  Here's a[n example of a new user created via texting Gambit, but they have a null SMS status.](http://aurora.dosomething.org/users/59f8e372a0bfad41c92ace4b)

aaronschachter:
Closes #202, Passing an email is no longer a requirement for creating Northstar users.

Also passes a default `sms_status`, it looks like users are being created without it.

cc @DFurnes


aaronschachter:
Verified for a new user: https://aurora-thor.dosomething.org/users/59f9f2de10707d043f244a85

DFurnes:
> Also passes a default `sms_status`, it looks like users are being created without it.

Yep, Northstar's got zero special logic here.

PR #210, opened 2017-11-01
Adds validate user middleware to Send Message

### LINKED ISSUE #205: POST /send-message: False alarm error rates
We'll be constantly getting `POST /send-message` requests from Blink for Users that don't have numbers, which means our error rates will always be false alarms in New Relic.

One idea is to throw a 422 and ignore 422 codes in New Relic, or throw a different obscure code and ignore that in New Relic.

Refs #189

### LINKED ISSUE #209: Check for User sms_status in POST /send-message
Need to test and confirm but I think we're missing a check when we load a User in `POST /send-message`.

Blink is sending all web signups to Gambit, to deliver an `externalSignupMenu`, checking if they have a valid phone number, and then sending. We need an additional check to make sure the User hasn't opted out of SMS messaging by inspecting the value saved for `sms_status`.

aaronschachter:
* Fixes #209 by checking User's subscription status
* Fixes #205 - throws a 422 if given User doesn't have a valid mobile US number

How to test:
* Send a stop keyword to update Northstar User's `sms_status`
* `POST /send-message` for your User, and an active Campaign with keywords
* Verify a 422 is returned

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -125,7 +125,11 @@ module.exports.sendErrorResponse = function (res, err) {
  */
 module.exports.sendResponseWithStatusCode = function (res, code = 200, message = 'OK') {
   const response = { message };
-  logger.debug('sendResponseWithStatusCode', { code, response });
+  if (code < 300) {
+    logger.debug('sendResponseWithStatusCode', { code, response });
+  } else {
+    logger.error('sendResponseWithStatusCode', { code, response });
+  }
 
   return res.status(code).send(response);
 };
--- a/lib/middleware/send-message/user-validate.js
+++ b/lib/middleware/send-message/user-validate.js
@@ -1,0 +1,34 @@
+'use strict';
+
+const helpers = require('../../helpers');
+const UnprocessibleEntityError = require('../../../app/exceptions/UnprocessibleEntityError');
+
+module.exports = function validateUser() {
+  return (req, res, next) => {
+    if (!req.user) {
+      return next();
+    }
+
+    if (req.user.sms_status === helpers.subscriptionStatusStopValue()) {
+      const error = new UnprocessibleEntityError('Northstar User is unsubscribed.');
+      return helpers.sendErrorResponse(res, error);
+    }
+
+    if (req.user.sms_paused) {
+      const error = new UnprocessibleEntityError('Northstar User conversation is paused.');
+      return helpers.sendErrorResponse(res, error);
+    }
+
+    try {
+      /**
+       * Sanity check E.164 format.
+       */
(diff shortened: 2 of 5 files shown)


lib/middleware/send-message/user-validate.js
```diff
@@ -0,0 +1,34 @@
+'use strict';
+
+const helpers = require('../../helpers');
+const UnprocessibleEntityError = require('../../../app/exceptions/UnprocessibleEntityError');
+
+module.exports = function validateUser() {
+  return (req, res, next) => {
+    if (!req.user) {
+      return next();
```
aaronschachter: This check is an example where this route and its middleware will get a lot cleaner when we split it out into 2: 

* sending a Support message from a Front agent
* sending a Campaign template from Blink / Rogue

lib/middleware/send-message/user-validate.js
```diff
@@ -0,0 +1,34 @@
...
+      return helpers.sendErrorResponse(res, error);
+    }
+
+    try {
+      /**
+       * Sanity check E.164 format.
+       */
+      req.platformUserId = helpers.formatMobileNumber(req.user.mobile);
+    } catch (err) {
+      const error = new UnprocessibleEntityError('Cannot format Northstar User mobile.');
```
rapala61: I think we need to suppress errors here for messages to users with no valid mobile #.

aaronschachter: @sergii-tkachenko -- I thought Blink always suppresses 422's, is that not the case? Could make sense to leave the explicitly suppress, but there could also be that (maybe very slight?) chance that we fix the user number before the retry count reaches the max...

rapala61: Mmm, maybe. This was a 500 error so I just suppressed when @sergii-tkachenko reported that it should be suppressed. I wasn't quite sure about Blink suppressing all 422s.

sergii-tkachenko: It always suppresses 422s. They show up in logs as warnings though.

sergii-tkachenko: You don't need to explicitly suppress 422s, it is a part of our retry contract. We need to write it down somewhere.

lib/helpers.js
```diff
@@ -125,7 +125,11 @@ module.exports.sendErrorResponse = function (res, err) {
  */
 module.exports.sendResponseWithStatusCode = function (res, code = 200, message = 'OK') {
   const response = { message };
-  logger.debug('sendResponseWithStatusCode', { code, response });
+  if (code < 300) {
```
sergii-tkachenko: Technically codes in range 300.399 are still considered successful. Example: `302 Found`.

PR #212, opened 2017-11-02
Fixes invalid logging, adds Northstar error check

aaronschachter:
* `logger.trace` is from winston, not heroku-logger

* The `try/catch` is fixing the Internal Server Error for me locally... just started seeing this error today per https://github.com/DoSomething/northstar/pull/680

### DIFF
--- a/lib/middleware/receive-message/user-update.js
+++ b/lib/middleware/receive-message/user-update.js
@@ -30,7 +30,7 @@ module.exports = function updateNorthstarUser() {
     }
 
     const data = {
-      last_messaged_at: req.inboundMessage.createdAt,
+      last_messaged_at: req.inboundMessage.createdAt.toISOString(),
       sms_paused: req.conversation.paused,
     };
     if (statusUpdate) {
@@ -51,6 +51,7 @@ module.exports = function updateNorthstarUser() {
       })
       .catch((err) => {
         let error = err;
+        // TODO: Move parsing this error into our Northstar JS library for all requests.
         if (err.response && err.response.body) {
           error = err.response.body.error;
         }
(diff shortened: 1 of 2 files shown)


lib/northstar.js
```diff
@@ -113,5 +113,7 @@ module.exports.fetchUserByMobile = function (mobile) {
  * @return {Promise}
  */
 module.exports.updateUser = function (userId, data) {
-  return exports.getClient().Users.update(userId, data);
+  return exports.getClient().Users.update(userId, data)
+    .then(res => res)
+    .catch(err => err);
```
rapala61: This is no bueno. It creates a false positive response up the Promise chain. Either don't catch the error or reject.

aaronschachter: Gotcha. This "fixed" receiving the Gambit reply, because it ignored the error we were getting for sending the Northstar update. I've updated the middleware to search for the specific Northstar error (and also fix the invalid `logger` call), but it'd be ideal to add this to all Northstar requests so don't get the "Internal Server Error" message back.

lib/middleware/receive-message/user-update.js
```diff
@@ -49,6 +49,12 @@ module.exports = function updateNorthstarUser() {
 
         return next();
       })
-      .catch(err => helpers.sendErrorResponse(res, err));
+      .catch((err) => {
+        let error = err;
+        if (err.response && err.response.body) {
```
rapala61: What do you think about surfacing `err.response.body` at the NS library directly, since it's a `superagent` idiosyncrasy? It's not merge blocking, but just food for thought for later.

aaronschachter: Definitely - that's what I meant about adding to all NS requests!

PR #213, opened 2017-11-03
Conversation middleware tests

aaronschachter:
Migrates some tests removed in https://github.com/DoSomething/gambit/pull/999/files#r148451514.

* Adds tests for the `conversation-get` and `conversation-create` middleware, following by example in https://github.com/DoSomething/gambit/pull/999.

    * I'm getting stuck on getting the `helpers.sendErrorResponse` stub to call, will comment inline

* Refactors `lib/twilio` with `createNewClient` and `getClient` functions to avoid requiring Twilio config vars in order test Conversation functions

    * Will deploy this to staging to confirm Twilio messages are successfully sent

### DIFF
--- a/lib/middleware/conversation-get.js
+++ b/lib/middleware/conversation-get.js
@@ -18,7 +18,7 @@ module.exports = function getConversation() {
         }
 
         logger.debug('getConversation', {
-          conversationId: conversation.id,
+          conversationId: conversation._id,
           topic: conversation.topic,
         });
 
--- a/lib/middleware/conversation-get.js
+++ b/lib/middleware/conversation-get.js
@@ -18,7 +18,7 @@ module.exports = function getConversation() {
         }
 
         logger.debug('getConversation', {
-          conversationId: conversation._id,
+          conversationId: conversation.id,
           topic: conversation.topic,
         });
 
--- a/lib/middleware/conversation-get.js
+++ b/lib/middleware/conversation-get.js
@@ -5,25 +5,23 @@ const Conversation = require('../../app/models/Conversation');
 const helpers = require('../helpers');
 
 module.exports = function getConversation() {
-  return (req, res, next) => {
-    Conversation.getFromReq(req)
-      .then((conversation) => {
-        if (!conversation) return next();
+  return (req, res, next) => Conversation.getFromReq(req)
+    .then((conversation) => {
+      if (!conversation) return next();
 
-        req.conversation = conversation;
-        const lastOutboundMessage = req.conversation.lastOutboundMessage;
-        if (lastOutboundMessage) {
-          req.lastOutboundTemplate = lastOutboundMessage.template;
-          req.lastOutboundBroadcastId = lastOutboundMessage.broadcastId;
-        }
+      req.conversation = conversation;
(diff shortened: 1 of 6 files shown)


test/lib/middleware/conversation-get.test.js
```diff
@@ -0,0 +1,97 @@
...
+  // setup
+  const next = sinon.stub();
+  sandbox.stub(Conversation, 'getFromReq').returns(conversationLookupFailStub);
+  sandbox.stub(helpers, 'sendErrorResponse').returns(sendErrorResponseStub);
+  const middleware = getConversation();
+
+  // test
+  await middleware(t.context.req, t.context.res, next);
+  // TODO: This doesn't get called in the test, but we're expecting it to :(
+  // helpers.sendErrorResponse.should.have.been.called;
```
aaronschachter: This returns `expected sendErrorResponse to have been called at least once, but it was never called'`.

One thing I notice in the debug output is that `AssertionError.proxy.rootObj.sendErrorResponse` is defined as circular:
```
     noCampaign: Function {},
        noReply: Function {},
        request: Object { … },
        rivescriptReply: Function {},
        sendErrorResponse: [Circular],
        sendResponseWithMessage: Function {},
        sendResponseWithStatusCode: Function {},
        slack: Object { … },
```

test/helpers/stubs.js
```diff
@@ -30,4 +34,44 @@ module.exports = {
...
     return req;
   },
+  getPlatform: function getPlatform() {
+    return 'sms';
+  },
+  getPlatformUserId: function getPlatformUserId() {
+    return mobileNumber;
+  },
+  middleware: {
+    getConversation: {
```
rapala61: This works, but it could be a good opportunity to start using the `factory` concept we use in Gambit Campaigns https://github.com/DoSomething/gambit/tree/master/test/utils/factories. It's more flexible since we can define different types of conversations we would like to create for testing.

test/lib/middleware/conversation-get.test.js
rapala61: I think it's never called because `sendErrorResponseStub` should be a rejected promise.

lib/middleware/conversation-get.js
```diff
@@ -5,25 +5,23 @@ const Conversation = require('../../app/models/Conversation');
 const helpers = require('../helpers');
 
 module.exports = function getConversation() {
-  return (req, res, next) => {
-    Conversation.getFromReq(req)
```
rapala61: Once I fixed the `Conversation` model stubbing, not returning the promise was another 🐛  in the middleware.

aaronschachter: Great catch! Should we make the same change in `createConversation` middleware ?

test/helpers/stubs.js
aaronschachter: Will add this in the next test / TGM migration cleanup, when we add tests removed in https://github.com/DoSomething/gambit/pull/1002

lib/middleware/conversation-get.js
rapala61: `createConversation` does return the promise, but ya, I think we should revise those tests in a future PR.

Issue #221, opened 2017-11-13
northstar.createNewClient RangeError: Maximum call stack size exceeded


aaronschachter:
This error is thrown on app starts when `lib/northstar` calls `createNewClient`. Doesn't seem to be affecting app functionality.

aaronschachter:
I am oddly not seeing this today.

aaronschachter:
👻

PR #222, opened 2017-11-13
Fixes logging in Northstar.createNewClient

aaronschachter:
Helps to debug errors like #221 - missed this from copy/paste from Gambit Campaigns (which uses `winston`, not the `heroku-logger` we're using here in Conversations)

### DIFF
--- a/lib/northstar.js
+++ b/lib/northstar.js
@@ -25,9 +25,9 @@ module.exports.createNewClient = function createNewClient() {
       baseURI: opts.baseUri,
       apiKey: opts.apiKey,
     });
-    logger.info(`${loggerMsg} success`, client);
+    logger.info(`${loggerMsg} success`, { client });
   } catch (err) {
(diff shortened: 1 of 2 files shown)


lib/northstar.js
```diff
@@ -25,9 +25,9 @@ module.exports.createNewClient = function createNewClient() {
       baseURI: opts.baseUri,
       apiKey: opts.apiKey,
     });
-    logger.info(`${loggerMsg} success`, client);
+    logger.info(`${loggerMsg} success`, { client });
```
rapala61: I think this one should be debug unless we want some specific info rather than the whole client object, because it potentially exposes credentials when logging. There is actually an info log I saw yesterday in production that exposes our Twilio Account and Token when creating a new client.

aaronschachter: Sometimes it's useful to set the log level to `debug` on prod, so I've just pushed up a commit that removes logging the client object both here and in `/lib/twilio`.

PR #228, opened 2017-11-15
Chore/224 add request id

### LINKED ISSUE #224: Add request_id to all requests in Conversations
# Story
As an O'dev, I want to add a request id to our kibana logs, so that I can trace all logged events that occurred processing that request.

# Spec
- [x] Some strategic logs should contain the `request_id` property to follow the path of the request in logstash.

rapala61:
# What does this PR do?
It injects the `request_id` into most logs that have access to the `req` or `res` objects, which contain the `request_id` metadata.

This solution doesn't add the `request_id` to **ALL** logs when processing a single request. This solution is a good compromise to update the logs already in place in very crucial spots, like middleware and helpers with access to `req` and `res` objects.

We can likely increase the usefulness of our logs if we continue to add them in all middleware in all routes and injecting the `request_id` on them.

# How to test?
- 👀 
- On staging, you can signup and reportback. You should see the `request_id` in many generated logs.
- Using Kibana, you can isolate a `request_id` and see a clear history of the steps taken to get a response. [Here is an example](http://logs.dosomething.org/app/kibana#/discover?_g=(refreshInterval:('$$hashKey':'object:2085',display:'10%20seconds',pause:!f,section:1,value:10000),time:(from:now-7d,mode:quick,to:now))&_a=(columns:!(message,conversationId,campaignId,text,signupId,userId,messageTemplate),filters:!(('$$hashKey':'object:790','$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:heroku_drain_id,negate:!f,value:d.1a141963-f3ea-42b0-b6bc-4a219b34adba),query:(match:(heroku_drain_id:(query:d.1a141963-f3ea-42b0-b6bc-4a219b34adba,type:phrase)))),('$$hashKey':'object:1254771','$state':(store:appState),meta:(alias:!n,disabled:!f,index:'filebeat-*',key:request_id,negate:!f,value:f7d3eaa8-16a9-4969-8b93-ddb0d980e599),query:(match:(request_id:(query:f7d3eaa8-16a9-4969-8b93-ddb0d980e599,type:phrase))))),index:'filebeat-*',interval:auto,query:(query_string:(analyze_wildcard:!t,query:'*')),sort:!('@timestamp',desc),vis:(aggs:!((params:(field:request_id,orderBy:'2',size:20),schema:segment,type:terms),(id:'2',schema:metric,type:count)),type:histogram))&indexPattern=filebeat-*&type=histogram).

# Relevant issues
Fixes #224


rapala61:
Example of how this logging can greatly improve troubleshooting:

screen shot 2017-11-14 at 11 44 47 pm

aaronschachter:
What do you think about adding a `lib/logger` that calls the `injectRequestId` instead of adding it into each middleware? If it was a separate module entirely, we could use it for https://github.com/DoSomething/gambit/issues/900 -- and track `request_id` between Conversations and Campaigns.

rapala61:
> We can likely increase the usefulness of our logs if we continue to add them in all middleware in all routes and injecting the request_id on them.

What I meant here was that, we should treat those places with access to the `req` and `res`  objects as crucial to our troubleshooting efforts, hence the need to add the `request_id` in all logs we use in these places, usually middleware. Some areas, like inside libraries, we don't have access to the `request_id` info, so we should log with `request_id` when we call them or receive the results.

> What do you think about adding a `lib/logger` that calls the `injectRequestId` instead of adding it into each middleware? If it was a separate module entirely, we could use it for https://github.com/DoSomething/gambit/issues/900 -- and track `request_id` between Conversations and Campaigns.

I think the way the `request_id` is [stored](https://github.com/DoSomething/gambit-conversations/pull/228/files#diff-d213b35d0a77db5706850392339bf8d2R23) from the beginning of a request is received in Convo, makes it a little unnecessary to add another library to recreate what a simple `heroku-logger` instance it's already doing for us. in order to associate the log to a `request_id`, I just have to extend the data passed as a second argument when logging, for [example](https://github.com/DoSomething/gambit-conversations/pull/228/files#diff-e33df83682f212a849fb570643a82748R237). I didn't want to manually do this, hence the request helper method `injectRequestId`.

I started down the path of a `lib/logger` but then realized that I would have to send it the `request_id` anyway, it would basically be a dummy module with just one method `injectRequestId` and it would return the same thing `helpers.request.injectRequestId` is doing, without doing any logging, but its called `logger`. Another alternative I also started implementing but decided not to, was to create a wrapper on top of `heroku-logger` and use that wrapper as our logger throughout Convo, but I feel like it's unnecessary given the simplicity of this solution.

I actually though about this a lot as you can see 😅. But I may be missing something anyway. If this context doesn't make sense and you have a better alternative, could you pseudo code a bit for me to get in the same page? Thanks!

rapala61:
> What do you think about adding a `lib/logger` that calls the `injectRequestId` instead of adding it into each middleware? If it was a separate module entirely, we could use it for https://github.com/DoSomething/gambit/issues/900 -- and track `request_id` between Conversations and Campaigns.

For https://github.com/DoSomething/gambit/issues/900 to receive a `request_id`, we need to make sure we add the `X-Request-Id` header to the POST request we do to Campaigns, then we need to parse it out and include in the `req` and/or `res` object.

rapala61:
@aaronschachter got it! Yeah, I like the idea. I thought of something similar, but it meant not using `heroku-logger` directly anywhere and use the wrapper instead, which I thought it could be a heavy handed approach, but if you are onboard, I can go that route. If I create this wrapper, do we want it to be an npm package to be used in both places? I can start the module here and create an issue to NPMify it later on.

aaronschachter:
👍 all around -- I'm on board -- I don't think of it as too as heavy handed. Definitely like the idea of Cam+Con sharing the same npm module. It's trippy and fun to think about sending the requestId to Northstar and Rogue too, to get the full trace on all systems...

Issue #206, opened 2017-11-01
Add New Relic custom parameters


aaronschachter:
Log the error message we return in an API response, as well as the usual suspects like `userId`, `campaignId`, `broadcastId`.

* [ ] Receive Message
* [ ] Import Message
* [ ] Send Message

Refs https://github.com/DoSomething/gambit/issues/834

aaronschachter:
Easiest way to implement might be to call the `newrelic.addCustomParameters` in our `helpers.sendResponse`, `helpers.sendErrorResponse` etc functions (vs adding each parameter in middleware) ?

aaronschachter:
If we go with ☝️ , we'd need to pass our `req` object into `helpers.sendErrorResponse` (and `helpers.sendResponseWithStatusCode`).

Another idea is creating[ `helpers.request`](https://github.com/DoSomething/gambit-conversations/blob/1.0.2/lib/helpers/request.js) setter functions that call the New Relic `addCustomParameters` as well as set the `req` variable.
e.g. `helpers.request.setUserId(userId)` , and making in the change in each relevant `/lib/middleware` function.

aaronschachter:
This feels safe to close out; can't think of any recent instances where it seemed like we were missing attributes in New Relic error reporting.

PR #235, opened 2017-11-22
Import Message Broadcast TLC - part 2

aaronschachter:
* Moves parsing functions defined in `parse-broadcast` into `contentful` to complete TODOs in #233 
* Adds more broadcast properties for #206  
* Adds tests for `lib/middleware/import-message/parse-broadcast`, `lib/helpers/analytics`

### DIFF
--- a/lib/middleware/import-message/parse-broadcast.js
+++ b/lib/middleware/import-message/parse-broadcast.js
@@ -1,36 +1,23 @@
 'use strict';
 
 const helpers = require('../../helpers');
+const contentful = require('../../contentful');
 const logger = require('../../logger');
-
-// TODO: Move inside lib/contentful
-function getCampaignId(broadcastObject) {
-  const campaign = broadcastObject.fields.campaign;
-  if (!campaign.fields) {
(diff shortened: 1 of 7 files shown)


lib/middleware/import-message/parse-broadcast.js
```diff
@@ -1,36 +1,21 @@
 'use strict';
 
 const helpers = require('../../helpers');
-const logger = require('../../logger');
-
-// TODO: Move inside lib/contentful
```
rapala61: Thanks for moving these!

test/helpers/stubs.js
```diff
@@ -35,6 +35,12 @@ module.exports = {
   getBroadcastId: function getBroadcastId() {
     return '72mon4jUeQOaokEIkQMaoa';
   },
+  getBroadcastMessageText: function getBroadcastMessageText() {
+    return 'Winter is coming, will you be prepared? Yes or No.';
```
rapala61: GOT? Yes!

PR #242, opened 2017-12-02
Macro helper

aaronschachter:
Branches off #241, continuing work on organizing the `lib/helpers` functions:
- finishes template helper cleanup (and renames as singular `helpers.template`)
- adds `lib/helpers/macro` and config


lib/helpers/template.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const logger = require('../../lib/logger');
+const config = require('../../config/lib/helpers/template');
+
+module.exports = {
+  getTextForTemplate: function getTextForTemplate(templateName) {
```
rapala61: I wonder if the "Template" naming  convention is unnecessary  now that the functions are part of a `tamplate` helper.

lib/middleware/receive-message/user-update.js
```diff
@@ -6,20 +6,17 @@ const northstar = require('../../northstar');
...
 module.exports = function updateNorthstarUser() {
   return (req, res, next) => {
-    // If we dont have a userId set, skip Northstar updates.
-    if (!req.userId) {
-      return next();
-    }
-
     const replyText = req.rivescriptReplyText;
+    // Note: We're setting SMS status for Slack users. If we ever integrate with another platform,
+    // we'll want to store separate platform subscription values on a Northstar User.
```
rapala61: ➕. Maybe we should update this step so that it only updates the user's `sms_status` on inbound SMS communication.

aaronschachter: That is what's happening? This middleware is only used in Receive Message.

lib/helpers/template.js
aaronschachter: I thought about this too... `isAskSignupTemplate` and `isGambitCampaignsTemplate` are tricky because we aren't checking specifically for a value match like we do the macro helper - e.g. `macro. isSupportRequestedMacro`. Going to merge for now but definitely open to renaming..

PR #248, opened 2017-12-08
added stop as the subscriptionStatusValue for stop

rapala61:
#### What does this PR do?
It changes the value of the stop subscription status from `undeliverable` to `stop`.

- Run `npm run all-tests`

#### Relevant Issues
- Pivotal #152612003

#### Tasks
- [x] Test on staging
- [x] User updated to sms_status: `stop` in Customer.io
- [x] User updated to sms_status: `stop` in Northstar.

### DIFF
--- a/lib/helpers/subscription.js
+++ b/lib/helpers/subscription.js
@@ -2,14 +2,15 @@
 
 const config = require('../../config/lib/helpers/subscription');
 
-function getStatusValueForKey(key) {
-  return config.subscriptionStatusValues[key];
+function getStatusForKey(key) {
+  return config.subscriptionStatuses[key];
 }
 
 module.exports = {
   statuses: {
-    active: () => getStatusValueForKey('active'),
-    less: () => getStatusValueForKey('less'),
-    stop: () => getStatusValueForKey('stop'),
+    active: () => getStatusForKey('active'),
+    less: () => getStatusForKey('less'),
+    stop: () => getStatusForKey('stop'),
+    undeliverable: () => getStatusForKey('undeliverable'),
   },
 };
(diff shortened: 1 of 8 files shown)


lib/middleware/receive-message/user-update.js
```diff
@@ -10,8 +10,8 @@ module.exports = function updateNorthstarUser() {
     // Note: We're setting SMS status for Slack users. If we ever integrate with another platform,
     // we'll want to store separate platform subscription values on a Northstar User.
     const currentStatus = req.user.sms_status;
-    const lessValue = helpers.subscriptionStatusLessValue();
-    const stopValue = helpers.subscriptionStatusStopValue();
+    const lessValue = helpers.subscription.statuses.less();
+    const stopValue = helpers.subscription.statuses.stop();
     let statusUpdate = null;
 
     if (helpers.macro.isSubscriptionStatusStop(replyText)) {
```
aaronschachter: @rapala61 This makes the check for `'stop'` -- but we'll need to additionally check for `'undeliverable'` for when updating a user's status  to`'active'`.

lib/helpers/subscription.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+const config = require('../../config/lib/helpers/subscription');
+
+function getStatusValueForKey(key) {
+  return config.subscriptionStatusValues[key];
+}
+
+module.exports = {
+  statuses: {
```
aaronschachter: Minor, but I wonder if it'd be more readable / consistent to either rename this as `statusValues` or rename ` config.subscriptionStatusValues` as `config.statues` ?

lib/middleware/receive-message/user-update.js
rapala61: Whoops, I was wondering if we needed to keep track of `undeliverable` as a status. Forgot to come back to that as I started writing the tests. Here's a 🌮 for that nice catch. Thanks.

rapala61: Updated.

PR #252, opened 2017-12-11
Broadcast aggregate totals + stats cache

### LINKED ISSUE #243: Expose Broadcast message counts via API
Refs https://www.pivotaltracker.com/story/show/153376335

Return Messages query count for a given `broadcastId`:

* [x]  Number of outbound 
* [x] Number of inbound

Inbound totals - blocked by #245 
* [ ] Number of Yes replies 
* [ ] Number of NO replies
* [ ] Number of Unsubscribe replies
* [ ] Number of Other replies

aaronschachter:
* Returns aggregate count of messages for a given `broadcastId`, grouped by the Message `direction` and `macro` properties. Closes #243 
    * Adds a compound index to `Messages`. For now, this index will be created upon deploy -- which is something Mongoose suggests [disabling](http://mongoosejs.com/docs/guide.html#indexes). I've been pondering using a config variable to control this vs. always manually creating... 

* Adds a `broadcastStats` instance of Cacheman to cache the formatted data to return in API response.
    * Splits the `lib/middleware/broadcasts-single/stats` middleware into `stats-cache-get` and `stats-cache-set`


test/lib/lib-helpers/broadcast.test.js
```diff
@@ -80,18 +73,63 @@ test('parseBroadcast should return an object', () => {
...
+  result.should.deep.equal(defaultStats);
+  broadcastHelper.parseMessageDirection.should.not.have.been.called;
+});
+
+test('formatStats should return object when array is passed', () => {
+  sandbox.spy(broadcastHelper, 'parseMessageDirection');
+  const inboundNoMacroCount = 43;
+  const inboundConfirmedCampaignMacroCount = 16;
+  const outboundCount = 205;
+  const aggregateResults = [
```
rapala61: We could use a [mock](https://github.com/DoSomething/gambit-conversations/pull/252/files#r156389161) here.

test/lib/lib-helpers/broadcast.test.js
```diff
@@ -80,18 +73,63 @@ test('parseBroadcast should return an object', () => {
...
+      count: inboundConfirmedCampaignMacroCount,
+    },
+    { _id: { direction: 'outbound-api-import' }, count: outboundCount },
+  ];
+  const result = broadcastHelper.formatStats(aggregateResults);
+  result.inbound.total.should.equal(inboundNoMacroCount + inboundConfirmedCampaignMacroCount);
+  result.inbound.macros.confirmedCampaign.should.equal(inboundConfirmedCampaignMacroCount);
+  result.outbound.total.should.equal(outboundCount);
+  // TODO: Why have you forsaken me? This should be called.
+  // broadcastHelper.parseMessageDirection.should.have.been.called;
```
rapala61: This is happening because `parseMessageDirection` is being stored in the `formatStats` scope, it's not accessing the exposed `module.exports.parseMessageDirection`. You would have to call it as `module.exports.parseMessageDirection` in https://github.com/DoSomething/gambit-conversations/pull/252/files#diff-7ccd6448aff97587fa8bdb77007f3c81R80. There are other tests in here that are assuming `parseMessageDirection` is being spied on, but in reality is not: Like [here](https://github.com/DoSomething/gambit-conversations/pull/252/files#diff-b309c91fe72283ac5533d45fb6ad0b16R99).

test/lib/lib-helpers/broadcast.test.js
```diff
@@ -80,18 +73,63 @@ test('parseBroadcast should return an object', () => {
...
+});
 
-  const result = broadcastHelper.getMessageCount(broadcastId, 'inbound');
-  result.should.equal(mockCount);
-  Message.where.should.have.been.called;
+test('formatStats should return default object when array without _id property is passed', () => {
+  sandbox.spy(broadcastHelper, 'parseMessageDirection');
```
rapala61: Not really being spied on because of https://github.com/DoSomething/gambit-conversations/pull/252/files#r156399452.

rapala61:
@aaronschachter Another point I forgot to mention before. We should move towards using Redis in Conversations. We don't really take advantage of caching with the current implementation (cacheman) because we have to warm up cache `#Dynos` times before all calls are "cached". It's possible to not hit a cached stat even if you refresh the page 16 times (the amount of Dynos now, ahead of the expected broadcast).

aaronschachter:
👍 re: Redis. I do think it'd be clean to move the Broadcast, Campaign Cacheman instances into the `cache` helper to consolidate -- and make it easy to switch out when we're ready.

Updated the tests, thanks for the 🦅 👀 !

PR #263, opened 2018-01-04
Update Northstar User address with Twilio data

### LINKED ISSUE #226: Store location information when updating Northstar User
Inbound Twilio messages include `FromCity`, `FromState`, `FromZip`, and `FromCountry` params.  Action items:
* [ ] For new users, Include this data when creating 
* [x] For existing users, check if we have address information. If not, update it when we're updating `last_messaged_at` in `lib/middleware/user-update`

```
2017-11-14T19:05:13.807133+00:00 app[web.7]: ToCountry=US ToState="" SmsMessageSid=SM3123de131e92040da527b711f8050bb4 NumMedia=0 ToCity="" FromZip=80621 SmsSid=SM3123de131e92040da527b711f8050bb4 FromState=CO SmsStatus=received FromCity="FORT LUPTON" Body=Stop FromCountry=US To=38383 MessagingServiceSid=MGc4401cd0c1729a8a7e7c66f50813b78a ToZip="" NumSegments=1 MessageSid=SM3123de131e92040da527b711f8050bb4 AccountSid=AC54a6683e7e5a2232aa9679003e84a685 From=[removed] ApiVersion=2010-04-01 level=debug message="POST /receive-message req.body"
```

aaronschachter:
#### What's this PR do?
Parses sender location info from an inbound Twilio request to `POST /receive-message`, and updates the corresponding Northstar User's address fields if they don't exist.

* Deploy this branch to staging (or run it locally)
* Clear out your Northstar User's address fields from Aurora
* Post to the relevant Gambit Staging / Dev Twilio numbers from your mobile number 
* Refresh your Northstar User's address in Aurora to verify address info has been saved

* This PR introduces a `lib/helpers/user` -- the function that determines whether to update a Northstar user's `sms_status` should be moved into it (along with tests)

* This is a little messy for new users: we could include the address data in our first User Create post (but instead pass it when we make a 2nd update request). Intentionally leaving this work out of the branch for now, as it's a cleanup task that may or may not be worth doing.

* A nice-to-have could be to modify the Consolebot to post the parameters of Twiilio request, to verify location info is saved without using Paw or mobile end-to-end testing.

#### Relevant tickets
Fixes #226 technically, but could warrant some cleanup to pass location data on User Create instead of Update for new users.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -5,7 +5,7 @@ module.exports = {
    * @param {object} user
    * @return {boolean}
    */
-  hasLocation: function hasLocation(user) {
+  hasAddress: function hasAddress(user) {
     if (user.addr_city && user.addr_state && user.addr_zip && user.country) {
       return true;
     }
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -1,0 +1,8 @@
+'use strict';
+
+module.exports = {
+  hasLocation: function hasLocation(user) {
+    const result = user.addr_city && user.addr_state && user.addr_zip && user.country;
+    return result;
+  },
+};
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -1,8 +1,14 @@
 'use strict';
 
 module.exports = {
+  /**
+   * @param {object} user
+   * @return {boolean}
+   */
   hasLocation: function hasLocation(user) {
-    const result = user.addr_city && user.addr_state && user.addr_zip && user.country;
-    return result;
+    if (user.addr_city && user.addr_state && user.addr_zip && user.country) {
+      return true;
+    }
+    return false;
   },
 };
--- a/lib/middleware/receive-message/user-update.js
+++ b/lib/middleware/receive-message/user-update.js
@@ -14,7 +14,7 @@ module.exports = function updateNorthstarUser() {
     };
 
     /**
-     * Check if User subscription status has changed.
+     * Check for User subscription status updates.
      */
     const replyText = req.rivescriptReplyText;
     // Note: We're setting SMS status for Slack users. If we ever integrate with another platform,
@@ -43,7 +43,7 @@ module.exports = function updateNorthstarUser() {
     }
 
     /**
-     * Check for User location.
+     * Check for User location updates.
      */
     if (req.platformUserLocation && !helpers.user.hasLocation(req.user)) {
       underscore.extend(data, req.platformUserLocation);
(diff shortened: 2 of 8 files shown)


lib/helpers/user.js
```diff
@@ -0,0 +1,14 @@
+'use strict';
+
+module.exports = {
+  /**
+   * @param {object} user
+   * @return {boolean}
+   */
+  hasAddress: function hasAddress(user) {
+    if (user.addr_city && user.addr_state && user.addr_zip && user.country) {
```
aaronschachter: @DFurnes -- Just wanted to run this by your 👀 -- if the Northstar User doesn't have *all* of these `addr` properties set, we'll update from the Twilio request. Wasn't sure about checking for `country`.

DFurnes: I think most people should have `country` since we set it from Fastly's geolocation header. But still makes sense to me to overwrite all of these here if not set.

rapala61: Should we also check for when the user's address has changed? Maybe just a lazy checking on the state could flag the need of updating the user's address. The downside is it could a little unpredictable for users that update their addresses manually on the website, the upside is we have a more accurate database of user's addresses.

rapala61:
@aaronschachter I tested on staging but my address didn't change in Aurora-thor. I made sure that the northstar user id was the same. The debug log didn't include the address properties: `data#last_messaged_at=2018-01-05T14:35:38.371Z data#sms_paused=false level=debug message="update userLocation"`.

lib/middleware/receive-message/user-update.js
```diff
@@ -28,15 +39,18 @@ module.exports = function updateNorthstarUser() {
...
-    };
     if (statusUpdate) {
       data.sms_status = statusUpdate;
     }
 
+    /**
+     * Check if we need to update User address.
+     */
+    if (req.platformUserAddress && !helpers.user.hasAddress(req.user)) {
+      underscore.extend(data, req.platformUserLocation);
```
rapala61: This should be `req.platformUserAddress `.

aaronschachter: Eek! I missed this when renaming the variables in 077a6a66b3aabb65c15ffd334f93d7faefc1cc1c. 🤦‍♂️  Will fix, verify, and reploy to staging. Sorry about that!

aaronschachter:
Updated and tested on staging.

lib/helpers/user.js
aaronschachter: There's a thread about updates in https://github.com/DoSomething/northstar/pull/695#issuecomment-355378153

For my mobile number, Twilio will always send my address as NJ because of the area code -- even though I live in CA now. We only want to posting a User's Twilio location to Northstar when the User doesn't have an address set; we're otherwise never touching the data in Gambit.

PR #268, opened 2018-01-08
Update User cleanup + tests

aaronschachter:
#### What's this PR do?
Completes the TODO in #263 for extracting `sms_status` updates into a helper function, and adds tests.

* Verify Northstar updates work as expected for:
    * Adding address if it doesn't exist via posting Twilio properties
     * Changing `sms_status` value via subscription keywords (e.g. `'stop'`, `'less'`)

Last TODO would be to add coverage and slightly refactor creating a User. Currently we create a User in the Create User middleware, posting to Northstar -- but we're also making a second post to Northstar to update the User's `sms_status`, `last_messaged_at`, etc. It's not the end of the world, but feels worth cleaning up? 

#### Relevant tickets 
#263 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -16,6 +16,17 @@ module.exports = {
   },
 
   /**
+   * @param {object} req
+   * @return {object}
+   */
+  getDefaultUpdatePayloadFromReq: function getDefaultUpdatePayloadFromReq(req) {
+    return {
+      last_messaged_at: req.inboundMessage.createdAt.toISOString(),
+      sms_paused: req.conversation.paused,
+    };
+  },
+
+  /**
    * @param {object} user
    * @param {string} rivescriptReplyText
    * @return {string}
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const macro = require('./macro');
+const statuses = require('./subscription').statuses;
+
 module.exports = {
   /**
    * @param {object} user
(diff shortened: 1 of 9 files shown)


lib/helpers/user.js
```diff
@@ -11,4 +14,43 @@ module.exports = {
...
+   * @return {string}
+   */
+  getSubscriptionStatusUpdate: function getSubscriptionStatusUpdate(user, rivescriptReplyText) {
+    const stopValue = statuses.stop();
+
+    if (macro.isSubscriptionStatusStop(rivescriptReplyText)) {
+      return stopValue;
+    }
+
+    if (macro.isSubscriptionStatusLess(rivescriptReplyText)) {
```
rapala61: is this a typo? `isSubscriptionStatusLess `. Seems like it should be `subscriptionStatusLess `

rapala61: Also for `isSubscriptionStatusStop` above.

aaronschachter: Not a typo -- https://github.com/DoSomething/gambit-conversations/blob/cleanup/update-status/lib/helpers/macro.js#L52

Seemed cleaner to have a helper function that returns `boolean` vs hardcoding the equality check, but I'm def open to discussion / changes if you have a cleaner approach in mind.

rapala61: Oh! sorry, I went by the changes in the commit and did not expand the code already there 🤦‍♂️ . It does feel like the equality helpers could be using the new `macros` functions, like:
```
module.exports.isSubscriptionStatusLess = function (text) {
  return (text === exports.macros.subscriptionStatusLess());
};
```

aaronschachter: Good call! I'll update and test/verify on staging.

aaronschachter: Updated and confirmed macros work as expected on staging.

PR #269, opened 2018-01-10
Create User cleanup + tests

aaronschachter:
#### What's this PR do?
* Modifies the User Update and Create middleware to avoid a double Northstar post for new users - one to create the user and a second to update their `sms_status` and address info. This is cleanup from #137

* Adds tests for `lib/northstar`

* Introduces `lib/middleware/receive-message/template-macro-reply` to DRY middleware for the Rivescript macros that need to return a hardcoded reply.

* Refactors `northstar.createUserForMobile` as `northstar.createUser`, deprecates the instance `conversation.createNorthstarUser` function for consistency with `user-update` #268 
    * Adds `helpers.user. getDefaultCreatePayloadFromReq` and `generatePassword`

* `POST /receive-message` with a Twilio payload, passing fields like `FromCountry`. Verify that only one Northstar post is made in console logs, and that the Northstar User has expected address fields set.

Needed before merge TODOs:
* [x]  Set `last_messaged_at` when creating a new User

Not for merge TODOs:
* Edge case create user subscription status should not be set to `'active'` (e.g. a new user texts STOP or LESS)
* Refactor Consolebot to post a Twilio payload, to verify Northstar address is created/updated

#### Relevant tickets
Fixes #263 

#### Checklist
- [x] Tested on staging.


test/lib/lib-helpers/user.test.js
```diff
@@ -23,11 +26,50 @@ const conversationFactory = require('../../helpers/factories/conversation');
...
+    .returns(mockDefaultPayload);
+  sandbox.stub(underscore, 'extend')
+    .returns(platformUserAddressStub);
+  sandbox.stub(userHelper, 'generatePassword')
+    .returns('taco');
+  const result = userHelper.getDefaultCreatePayloadFromReq(req);
+  result.source.should.equal(req.platform);
+  result.mobile.should.equal(req.platformUserId);
+  userHelper.getDefaultUpdatePayloadFromReq.should.have.been.calledWith(req);
+  underscore.extend.should.have.been.calledWith(mockDefaultPayload, req.platformUserAddress);
```
rapala61: A quick observation. This assertion could break later in a slightly different implementation for the same outcome. If tomorrow we wanted to use `Object.assign` instead of `underscore.extend` we would have to unnecessarily update this test.

PR #274, opened 2018-01-22
Fixes last outbound message loading bug

### LINKED ISSUE #219: Outbound message loading bug
# Bug
On retry, Convo attempts to load last inbound message and last outbound so to not create dupes of these when a request has failed and it's being retried. This mechanism assumes that an outbound message was created as part of the first attempt, but in some cases, the request fails before an outbound message is created and added as the `lastOutboundMessage` property of this user's Conversation.

#### Example: 
User with number `+15555555555` texted "Dream"
```
// Blink log
{"ToCountry":"US","ToState":"","SmsMessageSid":"SMXX","NumMedia":"0","ToCity":"","FromZip":"37216","SmsSid":"SMXX","FromState":"TN","SmsStatus":"received","FromCity":"NASHVILLE","Body":"Dream","FromCountry":"US","To":"SHORTCODE","MessagingServiceSid":"MGXX","ToZip":"","NumSegments":"1","MessageSid":"SMXX","AccountSid":"ACXX","From":"+15555555555","ApiVersion":"2010-04-01"}
```

The user had no messages in it's message history. The first message failed, because of an [H12](https://devcenter.heroku.com/articles/error-codes#h12-request-timeout). It created an inbound message but no outbound message. Notice the `metadata.retryCount`.

```
// Conversation, missing lastOutboundMessage
{
    "_id" : ObjectId("5a05aa08235b7900044d2058"),
    "updatedAt" : ISODate("2017-11-10T13:30:49.479Z"),
    "createdAt" : ISODate("2017-11-10T13:30:48.170Z"),
    "platformUserId" : "+15555555555",
    "platform" : "sms",
    "paused" : false,
    "topic" : "campaign",
    "__v" : 0,
    "campaignId" : 7927,
    "signupStatus" : "doing"
}
// Inbound message, no outbound message created
{
    "_id" : ObjectId("5a05aa08235b7900044d2059"),
    "updatedAt" : ISODate("2017-11-10T13:50:45.391Z"),
    "createdAt" : ISODate("2017-11-10T13:30:48.700Z"),
    "text" : "Dream ",
    "direction" : "inbound",
    "template" : null,
    "conversationId" : ObjectId("5a05aa08235b7900044d2058"),
    "topic" : "random",
    "platformMessageId" : "SMXX",
    "metadata" : {
        "requestId" : "17771be8-XXXX-4baf-XXXX-56a3e3de184d",
        "retryCount" : 527
    },
    "attachments" : [],
    "__v" : 0
}
```

Convo Heroku logs display the following error on each retry:
```
2017-11-10T13:48:40.321978+00:00 app[web.10]: code=500 response#message="Cannot read property 'text' of undefined" level=error message=sendResponseWithStatusCode
```
After some troubleshooting, I found the cause here: 
https://github.com/DoSomething/gambit-conversations/blob/master/app/models/Conversation.js#L291. Since`lastOutboundMessage` is undefined, attempting to get it's `text` property generates a 500 error.

https://github.com/DoSomething/gambit-conversations/blob/master/lib/helpers.js#L161. Because this request is being retried, this behavior will continue to occur, unless it's manually republished from Blink with a `retryCount` of 0. Giving it a chance to reprocess both the inbound and outbound. The problem with this workaround is that we have to manually delete the first instance of the inbound message that got stuck with a retryCount.

rapala61:
#### What's this PR do?
G-Con now checks if there was an outbound message created instead of assuming there is one, for requests that are being retried.

- 👀 
- Delete all messages that match your user's conversation id:
    - Run `db.messages.remove({conversationId: ObjectId(<CONVERSATION-ID>)})` in your local mongo cli runtime.
- Send a request with text `MENU` to `/receive-message` using Postman
    - Include the `x-blink-retry-count: 1` 
    - Include the `x-request-id: test` 
- You should not see any errors in your console or your Postman app. Previously, this would have triggered a `code=500 response#message="Cannot read property 'text' of undefined" level=error message=sendResponseWithStatusCode` error.
- Confirm both the `inbound` and `outbound` messages were created.

#### Relevant tickets
Fixes #219

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/config/lib/helpers/template.js
+++ b/config/lib/helpers/template.js
@@ -1,34 +1,47 @@
 'use strict';
 
+const underscore = require('underscore');
+
 const helpCenterUrl = 'http://doso.me/1jf4/291kep';
 // TODO: DRY menuCommand definition.
 // @see lib/helpers.js
 const menuCommand = 'menu';
 
+
+const templatesMap = {
+  campaignClosed: 'campaignClosed',
+  declinedSignup: 'declinedSignup',
+  declinedContinue: 'declinedContinue',
+  askContinueTemplates: {
+    askContinue: 'askContinue',
+    invalidAskContinueResponse: 'invalidAskContinueResponse',
+  },
+  askSignupTemplates: {
+    askSignup: 'askSignup',
+    invalidAskSignupResponse: 'invalidAskSignupResponse',
+  },
+  gambitCampaignsTemplates: {
+    askCaption: 'askCaption',
+    askPhoto: 'askPhoto',
+    askQuantity: 'askQuantity',
+    askWhyParticipated: 'askWhyParticipated',
+    completedMenu: 'completedMenu',
+    externalSignupMenu: 'externalSignupMenu',
+    gambitSignupMenu: 'gambitSignupMenu',
+    invalidCaption: 'invalidCaption',
+    invalidCompletedMenuCommand: 'invalidCompletedMenuCommand',
+    invalidPhoto: 'invalidPhoto',
+    invalidQuantity: 'invalidQuantity',
+    invalidSignupMenuCommand: 'invalidSignupMenuCommand',
+    invalidWhyParticipated: 'invalidWhyParticipated',
+  },
+};
+
 module.exports = {
-  askContinueTemplates: [
-    'askContinue',
-    'invalidAskContinueResponse',
-  ],
-  askSignupTemplates: [
-    'askSignup',
-    'invalidAskSignupResponse',
-  ],
-  gambitCampaignsTemplates: [
-    'askCaption',
-    'askPhoto',
-    'askQuantity',
-    'askWhyParticipated',
-    'completedMenu',
-    'externalSignupMenu',
-    'gambitSignupMenu',
-    'invalidCaption',
-    'invalidCompletedMenuCommand',
-    'invalidPhoto',
-    'invalidQuantity',
-    'invalidSignupMenuCommand',
-    'invalidWhyParticipated',
-  ],
+  templatesMap,
+  askContinueTemplates: underscore.values(templatesMap.askContinueTemplates),
+  askSignupTemplates: underscore.values(templatesMap.askSignupTemplates),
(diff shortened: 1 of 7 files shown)


rapala61:
@aaronschachter Just pushed part 2 - I'm still missing about 3 more tests. I'll ping when ready.

rapala61:
@aaronschachter this is ready for final 👀 . Also, it wouldn't hurt to go through at least these flows: signing up, reporting back, question, menu, info, stop, less, and bad words.

config/lib/helpers/template.js
```diff
@@ -1,42 +1,89 @@
...
-    subscriptionStatusLess: 'OK, great! We\'ll start sending you updates monthly instead of weekly! -Freddie, DoSomething.org.',
-    subscriptionStatusStop: `Ok, you'll stop getting texts from me. If you just want monthly texts, text LESS. Questions? ${helpCenterUrl}. If you wish to unsubscribe, do not respond to this message.`,
-    supportRequested: 'What\'s your question? I\'ll try my best to answer it.',
+
+const templatesMap = {
+  campaignClosed: 'campaignClosed',
+  declinedSignup: 'declinedSignup',
+  declinedContinue: 'declinedContinue',
+  rivescriptReply: 'rivescript',
+  memberSupport: 'memberSupport',
```
aaronschachter: I was confused why these properties exist for 5 templates and not all possible templates... it seems it's because these appear in neither gambitCampaignsTemplates or gambitConversationsTemplates?

The gambitCampaignsTemplates has historically only been used here to determine in receive-message middleware whether the last message sent to the User came from Gambit Campaigns. If it did, and we've made it thus far in the middleware (no other Rivescript triggers, Camaign keywords found) - we know to send the inbound text to Gambit Campaigns to determine response. Also why gambitCampaignsTemplates was an array, vs a property.

Could we add some documentation for why only these 5 are listed? Also definitely up for any other ideas if anything cleaner comes to mind...

rapala61: Yes! Sorry I just saw this. I will add documentation on it in a later PR.

PR #286, opened 2018-02-12
Remove v1 Send, Receive Message routes

aaronschachter:
#### What's this PR do?
Removes v1 endpoints, also seems to fix a bug causing requests to route to ` /api/v2/messages/api/v1/send-message`.

👁 

#### Checklist
- [x] Tested on staging.


app/routes/index.js
```diff
@@ -35,10 +35,6 @@ module.exports = function init(app) {
   app.use(parseMessageMetadataMiddleware());
 
   // v1
-  app.use('/api/v1/receive-message',
-    memberMessagesRoute);
-  app.use('/api/v1/send-message',
-    signupMessagesRoute);
```
aaronschachter: I've tried moving these to after to the v2 declarations, curious if we'd see a similar reversed (`POST /api/v2/messages/api/v1/receive-message`) route in New Relic. We do. Adding an [array of paths](https://stackoverflow.com/questions/15350025/express-js-single-routing-handler-for-multiple-routes-in-a-single-line) might work potentially, I didn't try -- since Blink is posting to v2 messages anyway (except for if we switch back to broadcasts v1), seems ok for removing these routes to be the fix.

PR #288, opened 2018-02-15
Refactoring Slack + adding Member Message params test

aaronschachter:
#### What's this PR do?
This is a breaking change for Slack integration that will get a corresponding PR in Gambit Slack. It aims to simplify the Conversations codebase, removing Slack features that we don't need to live on production (e.g. making a POST request to Slack API to deliver a message, when Slack is only used internally). The PR in Gambit Slack will take on responsibility of sending the Gambit reply to the Slack user.

* Refactors Consolebot to mock a Twilio request, for testing saving a User's address from a Twilio request

* Storing `slackChannel` as a means to deliver a Slack message was a bit brittle anyway. If we changed the Bot user for our Gambit Slack app, the Slack `channel` ID changes, breaking the experience for existing Slack Gambit conversations.

   * The Good: TODO - We can test Broadcasts and Signup messages via Slack by posting the `gambit-slack` as a `platform`. Gambit Slack will send the messages returned from the Conversations API response

   * The Bad: This breaks sending Front messages to a Slack conversation, as we're no longer saving a Slack channel. I think this tradeoff is OK - staff could check in Gambit Admin that their support message was sent, and they should be unpaused within Slack. 

It seemed like too much of another breaking change to add here, but I'd like to propose eventually refactoring Conversations to store a `userId` instead of a `platformUserId`. When we receive an inbound Twilio message, first get or create the User before getting or creating the Conversation for the `userId` and platform `'sms'`.

Verify replies work as expected over Twilio and Slack

#### Checklist
- [x] Tested on staging.


lib/helpers/request.js
```diff
@@ -30,12 +27,23 @@ module.exports = {
...
   },
+  setPlatform: function setPlatform(req, platform) {
+    req.platform = platform;
+    analytics.addParameters({ platform });
+  },
   setUserId: function setUserId(req, userId) {
     req.userId = userId;
     analytics.addParameters({ userId });
   },
+  suppressOutboundReply: function suppressOutboundReply(req) {
```
rapala61: Could we rename to `shouldSuppressOutboundReply` or `isSuppressOutboundReply` ? The way is named right now suggests that we are suppressing the outbound reply by calling this method. I had to come over and confirm implementation when I found the method being called somewhere else.

lib/middleware/messages/member/params.js
```diff
@@ -1,35 +1,34 @@
...
-    req.platformUserId = body.mobile;
-    req.platformMessageId = body.platformMessageId;
+    helpers.request.setPlatform(req, origin);
+    const body = req.body;
+    req.platformUserId = body.northstarId;
+    if (!req.platformUserId) {
+      const error = new UnprocessibleEntityError('Missing required northstarId.');
+      return helpers.sendErrorResponseWithSuppressHeaders(res, error);
+    }
+    req.platformMessageId = body.messageId;
```
rapala61: Minor, but I don't think is documented that we can send this property (`messageId`) to set a `platformMessageId` when using a `origin=:customPlatform`.

lib/middleware/messages/message-outbound-create.js
```diff
@@ -9,6 +9,9 @@ module.exports = function createOutboundMessage(config) {
       req.outboundTemplate, req)
     .then(() => {
       let promise = Promise.resolve();
+      // TODO: shouldPostToPlatform config is only needed for supporting v1 /import-message.
+      // Ideally this should DRY to use a req header, like how Consolebot is suppressing replies.
```
rapala61: A few concerns/Notes about this approach:
- It would work great for synchronous requests, from service `X` to G-Con. 
- It would be more complex with asynchronous requests, from service `X` to Blink to G-Con.
    - Headers are set inside the Blink worker. We would need to refactor how headers are handled in Blink.
    - Retries. We need to make sure the header is kept/regenerated on every retry. Otherwise, G-Con will attempt to load an outbound message or create if it doesn't find on retry.

test/lib/middleware/messages/member/params.test.js
```diff
@@ -0,0 +1,99 @@
...
+chai.use(sinonChai);
+
+// module to be tested
+const paramsMiddleware = require('../../../../../lib/middleware/messages/member/params');
+
+// sinon sandbox object
+const sandbox = sinon.sandbox.create();
+
+const stubParseTwilioBody = () => Promise.resolve(true);
+const stubOrigins = {
```
rapala61: Minor, not needed for merge. Sweet! Would be cool to add this to the stubs helper.

lib/middleware/messages/member/params.js
aaronschachter: 👍  - [Updated](https://github.com/DoSomething/gambit-conversations/blob/cleanup/slack/documentation/endpoints/messages.md#custom)!

lib/middleware/messages/message-outbound-create.js
aaronschachter: Updated this comment per our chat today -- doesn't seem to be a need for Blink to retry a Gambit request that doesn't post to messaging platform.

PR #291, opened 2018-02-20
Throw error in helpers.formatMobileNumber for invalid numbers

### LINKED ISSUE #290: Outbound Message Validate - Suppress Retry Bug for an invalid phone number
_BUG_
-------------------
#### Current Behavior

A string is returned in `helpers.formatMobileNumber` for User [5a8c855fa0bfad090c480e86](https://aurora.dosomething.org/users/5a8c855fa0bfad090c480e86), so the Validate Outbound Message isn't returning a 422. When we attempt to post this message to Twilio via `req.conversation.postLastOutboundMessageToPlatform()`, an error is returned but we're not suppressing retries, causing Blink to retry the message over and over.
The user's phone number contains 11 digits (not including `+1` prefix) instead of the expected 10.

> The 'To' number [removed] is not a valid phone number

#### Desired Behavior
The Validate Outbound Message middleware should have returned a 422 for this User ID.

#### Steps to Replicate
Post a Signup message for User `5a8c855fa0bfad090c480e86` (please make sure we didn't fix their profile though if you're going to do this)

#### Why This Matters
Unnecessary Blink retries == Unnecessary stress!

aaronschachter:
#### What's this PR do?
Adds a [google-libphonenumber](https://www.npmjs.com/package/google-libphonenumber) `isValidNumber` call to throw an error if the given mobile number is not valid. 

Doesn't add tests yet for sake of deploying this 

* Post a signup message with your User (who has a valid mobile), verify it's delivered
* Post a signup message for a User with invalid number (5a8c886b10707d0bde51084b), verify a 422 is returned

#### Relevant tickets
Fixes #290 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -90,7 +90,13 @@ module.exports.addBlinkSuppressHeaders = function addBlinkSuppressHeaders(res) {
 };
 
 module.exports.formatMobileNumber = function formatMobileNumber(mobile, format = 'E164', countryCode = 'US') {
+  logger.debug('formatMobileNumber params', { mobile });
   const phoneUtil = PhoneNumberUtil.getInstance();
   const phoneNumberObject = phoneUtil.parse(mobile, countryCode);
-  return phoneUtil.format(phoneNumberObject, PhoneNumberFormat[format]);
+  if (!phoneUtil.isValidNumber(phoneNumberObject)) {
+    throw Error('Invalid mobile number.');
+  }
+  const result = phoneUtil.format(phoneNumberObject, PhoneNumberFormat[format]);
+  logger.debug('formatMobileNumber return', { result });
+  return result;
 };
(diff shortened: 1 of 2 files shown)


rapala61:
I think it definitely fixes #290, but based on my comment above. We are also going to start getting requests caught in retry loops when the number is not reachable from our shortcode, for example.

aaronschachter:
Hmm, I see. 
> The 'To' phone number: [removed], is not currently reachable using the 'From' phone number: 38383 via SMS.

At this point, Gambit could potentially update northstar user status if we get one of these. Or consider posting the outbound message to Blink to post to Twilio... instead of posting to Twilio directly (but I think this may could get messy with retries potentially?)

aaronschachter:
What's odd is the [`The From phone number 38383 is not a valid, SMS-capable inbound phone number or short code for your account` error](https://www.twilio.com/docs/api/errors/21606) we're seeing occasionally.

`eaea0b35-7b34-4771-baa0-e782ed57ad3` is an example request where this occurs... we're seeing a bunch for this User (https://gambit-admin.herokuapp.com/conversations/5a8cd102c98a6c0004c89449) who seems to have signed up for 5 campaigns on the web (and is getting 4 failed retries for all of them).

PR #294, opened 2018-02-28
Removes POST /v1/import-message

aaronschachter:
#### What's this PR do?
Removes `POST /v1/import-message` route and related configs (e.g. toggle for `GET /v1/broadcasts/:id` webhook settings)

#### How should this be reviewed?
* Verify `GET /v1/broadcasts/:id` works as expected (returns `url` with `customerio-gambit-broadcast`, `body` with `northstarId` and `broadcastId` properties)

* Verify outbound messages work as expected (Broadcast, Front, Signup). No real changes here except removing the Message Outbound config `shouldPostToPlatform` property -- #293 introduces change to only post to platform for SMS Conversations.

* Verify `POST /v1/import-message` returns in a 404

#### Any background context you want to provide?
TODO:

* Update [wiki](https://github.com/dosomething/gambit-conversations/wiki) with v2 endpoints
* Deprecate Blink `customerio-sms-broadcast` resource

#### Relevant tickets
Closes first TODO in #293 

#### Checklist
- [x] Documentation added for new features/changed endpoints.
- [x] Tests added for new features/bug fixes.
- [x] Tested on staging.

PR #297, opened 2018-03-06
Conversation.userId [pr]

aaronschachter:
#### What's this PR do?
Adds a Conversation.userId property, deprecating the `platformUserId` property.

* Alters Twilio Member Message route to find or create Northstar User for given mobile number before getting/creating relevant Conversation. We're now storing Conversations by `userId` instead of `platformUserId` (what we'd use to save mobile numbers)

* Renames `req.platformUserId` as `req.userMobile` to make things a bit more readable. If we ever need to post messages to a different platform, like Facebook Messenger, we can save a separate `req.userFacebookId` property, or potentially revisit using the vaguer `req.platformUserId`.

* Refactors `Conversation.createAndPostOutboundReplyMessage` into separate functions, `createAndSetLastOutboundMessage` and `postLastOutboundMessage`.
    * Tweaks `createAndSetLastOutboundMessage` to update the Conversation.userId property if it doesn't exist (this would happen when we've already created a SMS Conversation for a given mobile number)
   * Tweaks `postLastOutboundMessage` to call `twilio.postMessage` with the `req.userMobile` instead of `req.platformUserId`

* Removes deprecated `import-message` check in Twilio helper per #294 

* Staging: verify that Twilio and Gambit Slack messaging work as expected upon a staging deploy. 
* Local: verify that adding `DS_CONSOLEBOT_SUPPRESS_REPLY=false` to your local `.env` results in expected behavior (success for legitimate mobile number, preferably your own, and failure for invalid mobile number)
 
Storing the Northstar User `userId` property is late to the game because the early stages of this project didn't rely on creating or getting a Northstar User. Since we've launched, we do rely on Northstar to store the User's subscription status (`user.sms_status`).  Saving the `userId` to Conversations and Messages makes the data team's life easier to query messages to find users who have sent messages that were caught bu certain Rivescript triggers, or to integrate with our [DS GraphQL](https://github.com/dosomething/graphql)

TODO (in a future PR):

* Add new diagrams in wiki with updated logic (check for User first before checking for a Conversation)

* DRY the Get User middleware -- this could be shared betweeen all routes, it just needs a config setting for whether to return 404 if User not found (only time we wouldn't send a 404 is for SMS, when we'd create a new User for the given mobile number)

* DRY setting default platform as `'sms'` if not found as a request body parameter.

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/155527887

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -4,7 +4,6 @@ const mongoose = require('mongoose');
 const logger = require('../../lib/logger');
 const Message = require('./Message');
 const helpers = require('../../lib/helpers');
-const twilio = require('../../lib/twilio');
 
 const campaignTopic = 'campaign';
 const defaultTopic = 'random';
@@ -306,14 +305,14 @@ conversationSchema.methods.createAndPostOutboundReplyMessage = function (text, t
     .then(() => {
       if (suppressReply) return Promise.resolve();
 
-      return this.postLastOutboundMessageToPlatform();
+      return this.postLastOutboundMessageToPlatform(req);
     });
 };
 
 /**
  * Posts the Last Outbound Message to Twilio for SMS conversations.
  */
-conversationSchema.methods.postLastOutboundMessageToPlatform = function () {
+conversationSchema.methods.postLastOutboundMessageToPlatform = function (req) {
   const messageText = this.lastOutboundMessage.text;
 
   // This could be blank for noReply templates.
@@ -321,7 +320,7 @@ conversationSchema.methods.postLastOutboundMessageToPlatform = function () {
     return Promise.resolve();
   }
 
-  return twilio.postMessage(this.platformUserId, messageText);
+  return helpers.user.sendTwilioMessage(req.user, messageText);
 };
 
 /**
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -19,7 +19,6 @@ const conversationSchema = new mongoose.Schema({
     index: true,
   },
   platform: String,
-  // platformUserId will one day be deprecated, was added before userId.
   platformUserId: {
     type: String,
     index: true,
@@ -68,9 +67,10 @@ conversationSchema.statics.getFromReq = function (req) {
       if (!helpers.request.isTwilio(req)) {
         return Promise.resolve(null);
       }
-      // Have we already saved a Conversation for this User by their mobile?
-      const queryByUserMobile = { platformUserId: req.userMobile };
-      return this.findOneAndPopulateLastOutboundMessage(queryByUserMobile, req);
+      // Have we already saved a Conversation for this User before we added userId?
+      // TODO: Remove this when all Conversations are be backfilled with userId.
+      const queryByPlatformUserId = { platformUserId: req.platformUserId };
+      return this.findOneAndPopulateLastOutboundMessage(queryByPlatformUserId, req);
     });
 };
 
@@ -189,6 +189,7 @@ conversationSchema.methods.getDefaultMessagePayload = function (text, template) 
     campaignId: this.campaignId,
     topic: this.topic,
     userId: this.userId,
+    platformUserId: this.platformUserId,
   };
   if (text) {
     data.text = text;
@@ -319,7 +320,7 @@ conversationSchema.methods.postLastOutboundMessageToPlatform = function (req) {
     return Promise.resolve();
   }
 
-  return twilio.postMessage(req.userMobile, messageText)
+  return twilio.postMessage(req.platformUserId, messageText)
     .then((twilioRes) => {
       // TODO: Store this metadata on our lastOutboundMessage:
       const sid = twilioRes.sid;
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -188,6 +188,7 @@ conversationSchema.methods.getDefaultMessagePayload = function (text, template) 
     conversationId: this,
     campaignId: this.campaignId,
     topic: this.topic,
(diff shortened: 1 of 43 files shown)


app/models/Conversation.js
```diff
@@ -266,50 +293,40 @@ conversationSchema.methods.setLastOutboundMessage = function (outboundMessage) {
...
-/**
- * @return {Promise}
- */
-conversationSchema.methods.getNorthstarUser = function () {
-  if (this.isSms()) {
-    return helpers.user.fetchByMobile(this.platformUserId);
-  }
-
-  return helpers.user.fetchById(this.platformUserId);
+  return twilio.postMessage(req.userMobile, messageText)
```
rapala61: Just for future feature request. Before sending the message. We could estimate amount of segments and store in metadata. Context for this request in this slack thread: https://dosomething.slack.com/archives/C2C8NLNAY/p1520350667000395.

lib/helpers/request.js
```diff
@@ -24,12 +24,27 @@ module.exports = {
     req.campaignId = campaignId;
     analytics.addCustomAttributes({ campaignId });
   },
+  setConversation: function setConversation(req, conversation) {
+    req.conversation = conversation;
+    analytics.addCustomAttributes({ conversationId: conversation.id });
+    const lastOutboundMessage = conversation.lastOutboundMessage;
```
rapala61: Minor, but probably more testable and readable if we split this function out and send a message as the argument to set the `req.lastOutboundTemplate` and `req.lastOutboundBroadcastId`.

lib/helpers/twilio.js
```diff
@@ -46,7 +37,7 @@ module.exports = {
           })
           .catch(error => reject(error));
       } else {
-        resolve('attachmentObject does\'t contain a redirectUrl');
+        resolve('attachmentObject doesn\'t contain a redirectUrl');
```
rapala61: 🔍 👍

test/app/models/conversation.test.js
```diff
@@ -68,17 +70,34 @@ test('createMessage should not call helpers.tag.render if direction is inbound',
...
+
+  await supportConversation.postLastOutboundMessageToPlatform(t.context.req);
+  twilio.postMessage.should.not.have.been.called;
 });
 
-test('getNorthstarUser should return northstar.fetchById for non SMS', async () => {
-  const result = await alexaConversation.getNorthstarUser();
-  helpers.user.fetchById.should.have.been.called;
-  result.should.equal(mockUser);
+test('postLastOutboundMessageToPlatform returns resolved Promise if conversation is not SMS', async (t) => {
```
rapala61: What about renaming to `postLastOutboundMessageToPlatform doesn't call twilio.postMessage if conversation is not SMS`

test/lib/lib-helpers/user.test.js
```diff
@@ -36,29 +40,60 @@ const platformUserAddressStub = {
...
   sandbox.stub(northstar, 'fetchUserByMobile')
     .returns(userLookupStub);
+
   const result = await userHelper.fetchByMobile(mockUser.mobile);
   northstar.fetchUserByMobile.should.have.been.called;
   result.should.deep.equal(userLookupStub);
 });
+
+// fetchFromReq
+test('fetchFromReq calls fetchById if !req.userMobile', async (t) => {
```
rapala61: Prob nitpicky, but test names could be a bit more friendly.

aaronschachter:
Thanks for the review! 

Re: breakage - this PR sets the Conversation `userId` when an outbound message is created, which should happen for any POST /messages request. I don't mind keeping it in though, I kept changing my mind about whether to call it `req.userMobile` or keep it as `platformUserId` (which feels a bit more cleaner / Gambit Adapterable for adding Messenger one day). 

It would definitely make sense to keep storing it on the Messages model, as phone numbers could potentially change Northstar Id's if two user accounts are merged.

app/models/Conversation.js
aaronschachter: It looks like we get a `num_segments` returned in a [Twilio StatusCallback](https://www.twilio.com/docs/guides/how-to-confirm-delivery), it could be nice to save via our future Blink StatusCallback request to Conversations `PUT v1/messages` (which I think should work for our Mongoose routes). We could update our Message metadata also to keep track of time message was queued, and delivered.

rapala61: Ah sweet beans. 👍

lib/helpers/request.js
aaronschachter: Added in 0736f9a30bd5c12be338ef0dcb821c25d0d826db

app/routes/messages/signup.js
```diff
@@ -21,6 +21,9 @@ const sendOutboundMessageMiddleware = require('../../../lib/middleware/messages/
 
 router.use(paramsMiddleware());
 
+// Validate Campaign Signup should send a Signup Menu message.
+router.use(getCampaignMiddleware());
```
aaronschachter: I noticed that Conversations were getting created in Staging for Signup Messages triggered by Ghost Inspector tests on Thor. There may be Conversations on production without any messages, because were created during a Signup message request that ends up failing later with a 422 because the Campaign is closed, or doesn't have any keywords. Need to investigate how many of these exist on production 🔍 , but this change should stop it 🚑

aaronschachter: Example: https://gambit-admin-staging.herokuapp.com/conversations/5a9eccee691c2400042be026.

PR #313, opened 2018-04-09
Twilio smart encoding + Gambit Campaigns API updates

aaronschachter:
#### What's this PR do?
* Makes name change per  https://github.com/DoSomething/gambit-campaigns/pull/1031, posting campaign activity to `/campaignActivity` instead of `/receive-message` Gambit Campaigns endpoint (identical routes, just a rename)
* Loads templates from the `botConfig` property instead of root campaign, to eventually deprecate `campaign.templates` in the Campaigns API
* Adds a `smartEncoded` parameter to encode outbound Twilio messages using Copilot's smart encoding per [Slack thread](https://dosomething.slack.com/archives/C5TSNJ6GL/p1523047665000246).

Verify campaign conversations work as expected.

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/gambit-campaigns.js
+++ b/lib/gambit-campaigns.js
@@ -144,7 +144,7 @@ function getCampaignActivityPayloadFromReq(req) {
 }
 
 /**
- * Posts data to the /receive-message endpoint.
+ * Posts campaign activity to Gambit Campaigns.
  */
 module.exports.postCampaignActivity = function (req) {
   const data = getCampaignActivityPayloadFromReq(req);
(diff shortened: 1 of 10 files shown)


lib/gambit-campaigns.js
```diff
@@ -146,13 +146,13 @@ function parseReceiveMessageRequest(req) {
 /**
  * Posts data to the /receive-message endpoint.
```
rapala61: Quick comment update .

aaronschachter:
Eek, great save! Will update (and add tests).

PR #325, opened 2018-05-10
Deprecating rivescript entries for defaultRivescriptTopicTrigger entries

aaronschachter:
#### What's this PR do?
This branch was an experiment in refactoring how we write the Rivescript that makes our chatbot reply. A Rivescript entry contains two fields:

* name - Single-value short text field for internal use, single value 
* rivescript - Single-value long text field to add freeform [Rivescript](https://www.rivescript.com/docs/tutorial)

Per our [spec](https://docs.google.com/document/d/1z5qLj2o-WTw4QDL0HlDMYoXQryvi_bfj_0IIuyOkkrY/edit#heading=h.ke7e6tx28a6f), the new `defaultRivescriptTopicTrigger`  entry contains two fields:

* `trigger` - Single-value text field to be used as a Rivescript trigger
* `response` - Single-value reference field to source the Rivescript reply, only allowing for content types:
    * `Message` - new content type with `text` and `attachments` fields
    *  `defaultRivescriptTopicTrigger` - this would be used for redirects, like how FORD should trigger the same message as JOIN does

This branch queries for all `defaultRivescriptTopicTrigger` entries instead of `rivescript` entries, and parses them as Rivescript triggers and replies via new `lib/helpers/rivescript`.

I've migrated the [Hello rivescript entry](https://app.contentful.com/spaces/owik07lyerdj/entries/i2a4NzCJsAoA2AekOSCmq) into two separate `defaultRivescriptTopicTrigger` entries:
* [hello](https://app.contentful.com/spaces/owik07lyerdj/entries/22UaWJW3AoUMG8QKA6Ym0) - with response set to a new Message entry
* [@hello [*] ](https://app.contentful.com/spaces/owik07lyerdj/entries/6fjKY5FNTycsyG2MeseMGg) - with response set to our "hello" `defaultRivescriptTopicTrigger` entry

I think this helps error-proof our Rivescript content, as well as making it easier to maintain:
screen shot 2018-05-10 at 10 37 23 am

Hello, hi, hey there, join, ford should all work locally but I haven't migrated any other entries yet. Wanted to make sure we nail down naming and the big picture before going to far into it.

* Eventually the `defaultRivescriptTopicTrigger.response` field will allow for postConfigs, to deprecate the `keyword` content type and support multiple postConfig conversations for a campaign.

* I like the name `defaultRivescriptTopicTrigger` because it is very descriptive but wow it is just way too long. I did consider just `defaultTopicTrigger`.... but naming a `rivescriptTopic` content type as `topic` just felt too general, so went with keeping the Rivescript in the name.

### DIFF
--- a/lib/helpers/rivescript.js
+++ b/lib/helpers/rivescript.js
@@ -1,0 +1,69 @@
+'use strict';
+
+const contentful = require('../contentful');
+const logger = require('../logger');
+
+function fetchDefaultRivescriptTopicTriggers() {
+  return new Promise((resolve, reject) => {
+    // TODO: fetch all rivescriptTopic entries to add any additional topics
+    contentful.fetchDefaultRivescriptTopicTriggers()
+      .then((triggers) => {
+        const result = triggers.map((trigger) => {
+          const rivescriptTrigger = module.exports.parseDefaultRivescriptTopicTrigger(trigger);
+          return rivescriptTrigger;
+        });
+        return resolve(result);
+      })
+      .catch(err => reject(err));
+  });
+}
+
+function parseDefaultRivescriptTopicTriggerResponse(contentfulEntry) {
+  if (contentful.isDefaultRivescriptTopicTrigger(contentfulEntry)) {
+    const triggerText = contentful.getTriggerFromDefaultRivescriptTopicTrigger(contentfulEntry);
+    return module.exports.formatTextAsRivescriptRedirect(triggerText);
+  }
+
+  if (contentful.isMessage(contentfulEntry)) {
+    const messageText = contentful.getTextFromMessage(contentfulEntry);
+    return module.exports.formatTextAsRivescriptReply(messageText);
+  }
+
+  throw new Error('Invalid content type for defaultRivescriptTopicTrigger.response');
+}
+
+function formatTextAsRivescriptRedirect(text) {
+  return `@ ${text}`;
+}
+
+function formatTextAsRivescriptReply(text) {
+  return `- ${text}`;
+}
+
+function formatTextAsRivescriptTrigger(text) {
+  return `+ ${text}`;
+}
+
+function formatRivescriptTrigger(triggerText, responseText) {
+  logger.debug('formatRivescriptTrigger', { triggerText, responseText });
+  const result = `${triggerText}\n${responseText}\n`;
+  return result;
+}
+
+function parseDefaultRivescriptTopicTrigger(trigger) {
+  const triggerText = contentful.getTriggerFromDefaultRivescriptTopicTrigger(trigger);
+  const rsTrigger = module.exports.formatTextAsRivescriptTrigger(triggerText);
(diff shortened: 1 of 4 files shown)


lib/helpers/rivescript.js
```diff
@@ -0,0 +1,69 @@
...
+function formatTextAsRivescriptRedirect(text) {
+  return `@ ${text}`;
+}
+
+function formatTextAsRivescriptReply(text) {
+  return `- ${text}`;
+}
+
+function formatTextAsRivescriptTrigger(text) {
+  return `+ ${text}`;
```
rapala61: I suggest we run a quick trim either here or at the `contentful.getTriggerFromDefaultRivescriptTopicTrigger` function. Also, just to be extra careful, we might want to find and replace instances of `@-+` at the beginning of the value sent to the `formatTextAsRivescript*` methods.

aaronschachter: Great call on trim ✂️ - making that update now . Contentful allows adding a regex for field validations, perhaps we can add the check there?

aaronschachter:
This branch has gone through another round of changes since the last review:

* Moves `module.exports` of the `lib/contentful` to the bottom of the file to fix linting errors
* Adds tests for `lib/helpers/rivescript` 
* Renames `brain/keywords.rive` -> `brain/macros.rive`, `brain/campaign.rive` ->`brain/topics.rive`
    * Migrates some macro triggers defined in Contentful into code, e.g. `stop @messaging [*]`

I've also been moving more triggers defined in the `rivescript` entries into `defaultRivescriptTopicTrigger` entries. Going to deploy this branch to begin testing on staging as I'm manually moving the rest over.

lib/helpers/rivescript.js
```diff
@@ -0,0 +1,108 @@
...
+
+/**
+ *  Returns a string to be used as a line of Rivescript.
+ *
+ * @param {String} operator
+ * @param {String} value
+ * @return {String}
+ */
+function formatRivescriptLine(operator, value) {
+  const rivescriptText = `${operator}${config.separators.command}${value}`.trim();
```
rapala61: What do you think of calling `.trim()` on the `value` itself? I think that is the variable that could potentially introduce a space at the beginning or at the end, since it's user input from Gambini.

lib/helpers/rivescript.js
```diff
@@ -0,0 +1,108 @@
...
+
+const contentful = require('../contentful');
+const logger = require('../logger');
+const config = require('../../config/lib/helpers/rivescript');
+
+/**
+ * Queries Contentful API for first page defaultRivescriptTopic entries and returns as an array of
+ * Rivescript triggers to be loaded into our Rivescript bot upon app start.
+ *
+ * TODO: Iterate through all pages in defaultRivescriptTopicResults instead of first page returned.
```
rapala61: Minor leftover of the old name `defaultRivescriptTopic` ?

aaronschachter:
I've migrated all content from the `rivescript` entries into code and new `defaultTopicTrigger` entries. Confirmed all looks well on staging.

PR #334, opened 2018-06-05
Feature/bertly integration

### LINKED ISSUE #330: Upgrade G-Conversations Node to 8.11.1
Node.js maintenance

rapala61:
#### What's this PR do?
It adds a library to detect links in outbound messages. Creates redirects for these links and replace them in place of the original links for outbound text.

Send yourself a broadcast that includes proper links (start with `http(s):`). The message should include shortened links instead of the original links.

1. ~Bertly seems to be throwing a 500 error when redirecting to a long URL right now. The redirect creation works however, so the message should still arrive to the mobile device.~ This was due to double encoding the URL sent to Bertly. Fixed in [commit](https://github.com/DoSomething/gambit-conversations/pull/334/commits/c6482d7fa062a5998b0d595aa8f79f7c2fe637b3).

2. Bertly returns the same short redirect URL when it receives a Link that has already been shortened.

#### Relevant issues
Fixes #330  

#### Checklist
- [x] Tested on staging.
- [x] Add ENV variables to Staging and Production:
    - DS_BERTLY_API_KEY=
    - DS_BERTLY_API_KEY_HEADER=
    - DS_BERTLY_API_BASEURI=
    - DS_BERTLY_API_ENABLED=


.nvmrc
```diff
@@ -1 +1 @@
-6.13.1
+8.11.1
```
aaronschachter: 🎉

lib/bertly/index.js
```diff
@@ -0,0 +1,95 @@
...
+});
+
+const logger = require('../logger');
+const restClient = require('./rest-client');
+
+/**
+ * textHasLinks
+ *
+ * @see http://markdown-it.github.io/linkify-it/doc/#LinkifyIt.prototype.test
+ * @param  {string} text description
```
aaronschachter: Minor, but I've been defining the params and returns with capital letters to try to keep consistent, e.g. `{String}`

PR #339, opened 2018-06-11
Change Topic macro [pr]

aaronschachter:
#### What's this PR do?

Gets the  `changeTopic` macro from #336 working, setting a `conversation.topic` to the Contentful Id `response` reference field of a`defaultTopicTrigger` (either a `textPostConfig`, `textPostConfig`, or `externalPostConfig`entry). This means once we're ready, we can deprecate the `keyword` content type (and related functions) by publishing `defaultTopicTrigger` entries for our current campaigns+topics. Keywords still work as is, handled by the `campaign-keyword` middleware.

* If a campaign keyword is sent, queries `GET /campaigns/:id` endpoint for reply content, using the first entry found in the campaign `topics` array property.

* If a `defaultTopicTrigger` executes a changeTopicMacro, queries `GET /topics/:id` endpoint for reply content, and sets the `conversation.campaignId` based on the topic `campaign` property.

    * Renames the `campaign-current` middleware as `topic-get` in `lib/middleware/messages/member`<p>

* Renames functions like `replies.continueCampaign` and macros like `confirmedCampaign` as `continueTopic`, `confirmedTopic`

* Renames the `campaignMenu` macro as just `menu`, in case we ever want to pitch topics without campaigns, or profile update topics (e.g. #340)

* Renames the `gambit` macro we use for our [catch-all trigger in Rivescript](https://www.rivescript.com/docs/tutorial#catch-all-trigger) as `catchAll`, to improve readability. The `gambit` macro name is a relic from when this app was a MVP named  [`slothie`](https://github.com/DoSomething/slothie)
 and was posting to [`gambit`](https://github.com/DoSomething/gambit).

    * Consolidates the `campaign-closed`, `template-no-campaign`, `parse-ask-signup-response`, `parse-ask-continue-response`, and `campaign-continue` middleware into a `macro-catch-all` middleware. I was a little conflicted about merging them, but each one is simple enough it made it feel more confusing to know which order we check the different conversation states in. Curious if you think this helps or hurts simplification?

* Verify that keyword entries, e.g. `'mirror'`, work as expected, as we can't unpublish them before this is deployed to production.
    * Topic should be set to the first topic found in the `GET /campaigns/:id` response for a campaign

* Verify that `defaultTopicTrigger` entries with topics set work as expected:
    * Run `gambit-campaigns` in Preview mode and this branch of `gambit-conversations` 
    * Verify draft `defaultTopicTrigger` entries change the conversation topic and campaign as expected, e.g. `spit` should change topic to `tv7e98JGXmMM2kskGaUA2` and campaign to 2178.

* This doesn't update the broadcast or signup message routes to set the topic Id as topic, but they do work. If we send a broadcast this week, it will be setting the rivescript topic to a [hardcoded one](https://github.com/DoSomething/gambit-conversations/blob/2.8.1/brain/topics.rive), so we don't need to worry about setting a contentful Id as a `topic` 

* This needs way more unit tests, but I wanted to open at this point to make it readable.

#### Relevant tickets

Ground work for:

* `POST /v2/messages?origin=afterSignupReminder` endpoint that will check a user's conversation topic and send a reminder template message if it exists for the user's current topic: https://www.pivotaltracker.com/story/show/156174270

* Refactoring the Broadcast content type to reference a topic entry instead of a `campaign` entry: https://www.pivotaltracker.com/story/show/157369418

* A Typeform topic, where a campaign may be optional: https://www.pivotaltracker.com/story/show/157642858

#### Checklist
- [x] Tested on staging.


app/models/Conversation.js
```diff
@@ -84,7 +84,20 @@ conversationSchema.statics.findOneAndPopulateLastOutboundMessage = function (que
...
 /**
- * Set topic property.
+ * Updates topic and campaign per given topic object.
+ * @param {Object} topic
+ * @return {Promise}
+ */
+conversationSchema.methods.changeTopic = function (topicObject) {
+  this.topic = topicObject.id;
+  this.campaignId = topicObject.campaign ? topicObject.campaign.id : null;
+  logger.debug('conversation.changeTopic', { topic: this.topic, campaignId: this.capmaignId });
```
rapala61: Minor Typo.

lib/helpers/request.js
```diff
@@ -1,11 +1,183 @@
...
+  // Inject a campaign property into the topic, as we'll need it for the call to changeTopic.
+  topic.campaign = campaign;
+  return module.exports.changeTopic(req, topic);
+}
+
+/**
+ * @param {Object} req
+ * @return {Promise}
+ */
+function executeChangeTopicMacro(req) {
```
rapala61: Q: Now that we have `defaultTopicTriggers` that work as redirects for other `defaultTopicTriggers` that ultimately point to the topic. Would it be worth it to check for the edge-case where the use is already in that topic, but texted the redirect trigger? This way we save a round trip to G-Campaigns? Example: Texting `safety`, then texting `safe` after.

aaronschachter: We don't do this currently for keywords: if a user texts in a keyword for a campaign they're already in, we still query the Content API and set them accordingly. 

This does sound like a reasonable check to make, but don't think we need to get it in for merge+deploy as it's a rarity. I'd be happy to add the check into a cleanup branch I'm working on which branches off this one, updating he broadcast and signup middleware per these changes.

rapala61:
One thing I noticed is that the `inbound` message's `topic` and `campaignId` differ from the `outbound-reply` message, but share the same `match` and `macro`. This makes the data in the `inbound` message a little ambiguous.

Example:
```
{
  "_id": ObjectId("5b200abdc3b655db526d7bc1"),
  "updatedAt": ISODate("2018-06-12T18:02:37.518Z"),
  "createdAt": ISODate("2018-06-12T18:02:37.518Z"),
  "text": "One of the most impactful ways to create change in our communities and on issues we care about is by making our voices heard through voting. Your generation has the power to decide the next election. \n\nWhether you are old enough to vote or not -- you can make a difference. Tag a friend (who is 18 or older) on Facebook and give them an easy way to register to vote. By sharing this, you'll be entered to win a 2000 scholarship: https://dosome.click/<redacted>",
  "direction": "outbound-reply",
  "template": "startExternalPost",
  "conversationId": ObjectId("5b1fe952bcaa43dac351f1c8"),
  "campaignId": 7059,
  "topic": "5MSUDKlVp6kqkUMw8gW004",
  "userId": "<redacted>",
  "broadcastId": null,
  "match": "vcard",
  "macro": "changeTopicTo5MSUDKlVp6kqkUMw8gW004",
  "metadata": {
    "requestId": "eaf546d8-3b19-4701-849b-de43b23a76af"
  },
  "attachments": [ ],
  "__v": 0
}
{
  "_id": ObjectId("5b200ab9c3b655db526d7bc0"),
  "updatedAt": ISODate("2018-06-12T18:02:33.003Z"),
  "createdAt": ISODate("2018-06-12T18:02:33.003Z"),
  "text": "vcard",
  "direction": "inbound",
  "template": null,
  "conversationId": ObjectId("5b1fe952bcaa43dac351f1c8"),
  "campaignId": 7951,
  "topic": "5xa3oDEx4Ao0Sm2qoQCeI",
  "userId": "<redacted>",
  "broadcastId": null,
  "match": "vcard",
  "macro": "changeTopicTo5MSUDKlVp6kqkUMw8gW004",
  "metadata": {
    "requestId": "eaf546d8-3b19-4701-849b-de43b23a76af"
  },
  "attachments": [ ],
  "__v": 0
}
```

rapala61:
Just an observation. Sorry if this is the way it has worked forever. Ignore me if so. 😬

aaronschachter:
It is the way it's always worked, and I do agree it can seem a little confusing to have them in both messages... it's also required though to be able to query both inbound and outbound based on the match. This is an instance of where I could see a use case for a separate `request` entity based on the `requestId` that stores the Rivescript response for the inbound message (setting the `macro` and `match`)

PR #344, opened 2018-06-15
unsubscribed + askSubscriptionStatus topics

### LINKED ISSUE #340: subscriptionStatusActive macro + Update Frequency topic
_FEATURE OVERVIEW_
-------------------
#### User Story
As an admin, I want Gambit support for:

* a `'weekly'` default topic trigger that executes a `subscriptionStatusActive` to update the user's Northstar `sms_status` as `'active'` 

* An `'update_frequency'` Rivescript topic to be used for a broadcast, asking users to set their broadcast frequency preferences by asking a multiple choice question. Each option should execute either the `subscriptionStatusActive`, `subscriptionStatusLess`, or `subscriptionStatusStop` macros - copy TBD

#### Additional Information
* https://dosomething.slack.com/archives/C2C8NLNAY/p1528734685000471

#### Why This Matters

* Allows us to send a broadcast to verify user's broadcast frequency preferences
* Gives members the option to change to '`active'` if their status is currently `'less'`.

#### Current Workaround w/o Feature (if applicable)
Custom Rivescript topic that doesn't update user status, data team queries Gambit DB for broadcast responses and updates users' `sms_status` manually

aaronschachter:
#### What's this PR do?

Adds 2 additional subscriptionStatus macros (`subscriptionStatusActive`, `subscriptionStatusResubscribed`) and 2 hardcoded topics:

* `askSubscriptionStatus` - Prompts user to execute either a `subscriptionStatusActive`,  `subscriptionStatusLess`, or `subscriptionStatusStop` macro. This topic can be joined by texting a STATUS keyword for now, it's also the topic we'll use for our permissioning broadcasts.

* `unsubscribed` - When a `subscriptionStatusStop` macro fires, it changes the conversation topic to this new topic, where only INFO, HELP, LESS, and JOIN are defined as triggers. All other user messages will not receive a response if the user's conversation topic is set to `unsubscribed` (creating an blank message with template `noReply`).
    * We may change this to a new `subscriptionStatusUnsubscribed` template that prompts user to text JOIN, but keeping as `noReply` per spec

Also adds support to change topics when calling macros via Rivescript syntax, slightly refactoring work done in #339 to remove `getTopicIdFromChangeTopicMacro` function

Text STATUS and test various options to verify your Northstar User `sms_status` and Gambit Conversation `topic` update accordingly.

A new Conversation has a default value of `'random'` - we should create a new `'unconfirmed` or `anonymous` topic to use for when an unrecognized mobile number sends a message that isn't a reserved JOIN keyword (e.g. JOIN, campaign signup keywords)

#### Relevant tickets
Fixes #340

#### Checklist
- [x] Tested on staging.
- [x] Pinged a PM if this is a larger PR that would benefit from some additional testing love.


brain/topics.rive
```diff
@@ -19,6 +39,35 @@
...
+> topic askSubscriptionStatus includes random
+
++ a
+- subscriptionStatusActive{topic=random}
+
++ b
+- subscriptionStatusLess{topic=random}
+
++ c
+- DoSomething is a community of 6 million young people taking action in their community. Read more about what we do: https://www.dosomething.org/us/about/who-we-are. Do you have a question for me? Text QUESTION and I’ll reply within 24 hours!\n\nDo you want to hear from us A) Weekly B) Monthly or STOP
```
aaronschachter: @DFurnes @rapala61 - Heads up that this copy contains a link, which would get Berlied. Our outbound to the full list won't contain a list, but we will for sure see a spike if users text back the "more info" option

rapala61:
One thing to note: When sending a picture, I got a `ERR: No Reply Matched` to my phone.

aaronschachter:
Really good catch @rapala61. I can replicate this while unsubscribed... and also caught a (minor IMO) bug that a user can break out of the `askSubscriptionStatus` topic when sending through an image. I'll need to investigate how we can catch these but don't think it should block sending out the broadcast. cc @ngjo @classicfreddie

aaronschachter:
It currently breaks them out of the `askSubscriptionStatus` topic and prompts whether you wish to continue (you can an example[ on my Gambit Admin profile](https://gambit-admin-staging.herokuapp.com/users/5547be89469c64ec7d8b518d)) for my reply: https://gambit-admin-staging.herokuapp.com/requests/b0d6b22b-037d-4702-835b-fdb849fdaf0e

Also, while we're chatting here -- there are two features that have come up in meetings and Slack not addressed in this PR:

1 - Sending a new template named `subscriptionStatusUnsubcribed` with copy like  "These are DoSomething alerts, 4 msg/mo Text JOIN to subscribe STOP to oputout" instead of the silent `noReply` we currently sending 
2 - Updating a user's sms_status to `'unknown'`.  Anyone who answers C Need More Info would be updated with this status, and potentially anyone who exists broadcast without answering (by sending an image)... although higher priority there would be to prevent images from changing the conversatino topic.

ngjo:
Also, closing the loop @aaronschachter -- not a blocker for sending back an image takes them outta the flow. I'll create an icebox card for that!

PR #345, opened 2018-06-18
Don't resubscribe unless subscriber macro triggered

aaronschachter:
#### What's this PR do?

Fixes bug in #344 -- sending Gambit a subsequent message after sending a STOP request would silently set the user's status back to active.

* Also adds our [`@unsubscribe` substitution](https://github.com/DoSomething/gambit-conversations/blob/2.10.0/brain/index.rive#L92) as a trigger to return the `subscriptionStatusStop` value. A todo here could be to return `null` if the subscriptionStatusMacro triggered is equal to the user's current status. 

* Verify your user is sms subscribed (status `active` or `less`)
* Send a STOP message to Gambit, verify user status is `stop`
* Send random message to Gambit, verify user status is still set to `stop`
* Send JOIN message to Gambit, verify user status is` active`, repeat all steps with LESS message to verify status is saved as `less`

Would be nice to start up on more integration tests for 🐛 like these 🤦‍♂️ 💻 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -97,14 +97,20 @@ module.exports = {
    * @return {string}
    */
   getSubscriptionStatusUpdate: function getSubscriptionStatusUpdate(user, rivescriptReplyText) {
-    const activeStatus = statuses.active();
+    const activeStatusValue = statuses.active();
+    // Note: We're checking SMS status for Slack users. If we ever integrate with another platform,
+    // we'll want to store separate platform subscription values on a Northstar User.
+    const currentSubscriptionStatus = user.sms_status;
 
+    /**
+     * Check if rivescriptReplyText is a subscriptionStatus macro.
+     */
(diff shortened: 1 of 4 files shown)


lib/helpers/user.js
```diff
@@ -115,23 +121,20 @@ module.exports = {
...
-    // (with a non subscription status macro), subscribe them.
-    //
-    // TODO: Instead of blindly accepting rivescriptReplyText as a subscription at this point,
-    // validate that rivescriptReplyText is a reserved JOIN or campaign signup trigger.
-    //
-    if (!this.isSubscriber(user)) {
-      return activeStatus;
+    // TODO: Once we deprecate keywords, check if rivescriptReplyText is a changeTopic macro,
+    // and return active status value.
+    // @see https://www.pivotaltracker.com/story/show/158444719
```
rapala61: This comment is pointing to a deleted story.

PR #347, opened 2018-06-19
Set user sms_status to pending if their topic changes to askSubscriptionStatus

aaronschachter:
#### What's this PR do?

Adds logic to the `Conversation.setTopic` function to update a user's `sms_status` to `pending` if the topic has been changed to `askSubscriptionStatus`.

* ~~We're waiting on `pending` to be added as a valid `sms_status` option, so this branch sets the status to `unknown` to unblock development and testing. See https://www.pivotaltracker.com/n/projects/2019429/stories/158439520~~

* Adds `helpers.user.setPendingSubscriptionStatusForUserId` to execute the update request to Northstar for given `userId` param

I've set up a test broadcast, `5pcoF2S9ZC68qqQygCCEqG`, with topic `askSubscriptionStatus`. Send yourself the broadcast and verify your Northstar user's status is set to `pending`.

screen shot 2018-06-19 at 10 15 18 am

* The `setSupportTopic`, `setCampaignTopic` functions could update NS status as well, but we aim to deprecate both of these topics:

    * No longer use a support topic per https://www.pivotaltracker.com/story/show/157987170
    * A campaign broadcast should set the campaign topic instead of a campaign -- this is one of the last remaining cleanup tasks from the topic work that has been on hold while we drop everything to handle compliance issues.

#### Checklist
- [x] Tested on staging.
- [x] Pinged a PM if this is a larger PR that would benefit from some additional testing love.

### DIFF
--- a/test/unit/lib/lib-helpers/user.test.js
+++ b/test/unit/lib/lib-helpers/user.test.js
@@ -199,3 +199,14 @@ test('isPaused should return user.sms_paused', (t) => {
   const result = userHelper.isPaused(user);
   t.deepEqual(result, user.sms_paused);
 });
+
+// setPendingSubscriptionStatusForUserId
+test('setPendingSubscriptionStatusForUserId should call northstar.updateUser with pending status', async () => {
+  const userId = mockUser.id;
+  sandbox.stub(northstar, 'updateUser')
+    .returns(Promise.resolve({}));
+
+  await userHelper.setPendingSubscriptionStatusForUserId(userId);
+  northstar.updateUser.should.have.been
+    .calledWith(userId, { sms_status: subscriptionHelper.statuses.pending() });
+});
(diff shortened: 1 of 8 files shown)


test/unit/lib/lib-helpers/user.test.js
```diff
@@ -199,3 +199,14 @@ test('isPaused should return user.sms_paused', (t) => {
   const result = userHelper.isPaused(user);
   t.deepEqual(result, user.sms_paused);
 });
+
+// setPendingSubscriptionStatusForUserId
+test('setPendingSubscriptionStatusForUserId should call northstar.updateUser with pending status', async () => {
```
rapala61: Is it possible to also test the case of NS being down and what errors are triggered and what not? Given this broadcast is going to be so critical.

aaronschachter: As per our chat, will look into adding more tests in a subsqequent PR. I'd like to merge this, then merge #348 - then make final copy updates to `topics.rive` and add tests if there's time.

Issue #342, opened 2018-06-14
Back fill all production messages to include `userId`.


rapala61:
_BUG_
-------------------
#### Current Behavior
- some messages do not contain the `userId` property when the conversation does.

#### Desired Behavior
- if we are storing the userId in the message we should keep it consistent on all messages. 

#### Steps to Replicate
These 2 messages for example:
```
/* 48 */
{
    "_id" : ObjectId("5b108c80c6fcef0004c14d4c"),
    "direction" : "outbound-api-send",
    "conversationId" : ObjectId("5a0601f4c68d250004f0c994")
}

/* 49 */
{
    "_id" : ObjectId("5b108c8011dd1a0004f4d4d0"),
    "direction" : "outbound-api-send",
    "conversationId" : ObjectId("5a0601f0ad7de60004f77599"),
    "userId" : "55a75707469c649d2b8cecf9"
}
```
#### Why This Matters
It makes it precarious for complex querying that some messages include it and others don't.

aaronschachter:
We have an existing Pivotal story for this issue: https://www.pivotaltracker.com/story/show/155527953

aaronschachter:
Technically could be a separate issue, but we should also backfill any `conversation` documents that don't contain a `userId`, as that property was added a few months after the first Conversations release.

Refs https://dosomething.slack.com/archives/C2C8NLNAY/p1529512662000400?thread_ts=1529506973.000763&cid=C2C8NLNAY

rapala61:
This elder 🐛 made an appearance while testing the member's deletion feature in Gambit. [Conversation](https://dosomething.slack.com/archives/C1V0M6RPE/p1566316182003100?thread_ts=1566315070.001000&cid=C1V0M6RPE).

@ngjo Here's the issue for this 🐛. Here's the [card](https://www.pivotaltracker.com/story/show/168031609) I just created in the Platform's Pivotal board.

PR #352, opened 2018-06-20
Backfill conversation.userId upon finding existing conversation

aaronschachter:
#### What's this PR do?

Moves the backfilling of a `conversation.userId` property out of the `Conversation.createAndSetLastOutboundMessage` instance method, and instead into the `Conversation.findOneAndPopulateLastOutboundMessage` static method instead.

#351 fixed a bug in #347, stopping errors from being thrown when a conversation didn't have a `userId` set upon receiving a `askSubscriptionStatus` broadcast by skipping the user update completely. This PR moves the backfill into the `conversation-get` middleware, which means we'll have a backfilled `userId` property available by the time we're ready to set the conversation topic (and update user accordingly if our topic change is to `askSubscriptionStatus`)

This bug occurred because it's the first time we've had to make an update to the user when sending a broadcast message (and because we're sending to our full list, where some users who haven't sent or received messages in a long time may not have conversations with userId's).

Test sending yourself an `askSubscriptionStatus` broadcast  with the following steps:

* Make sure your `sms_status` is set to `'active'` or `'less`' (or null could work too). Also make sure your conversation topic is not set to `askSubscriptionStatus`, as the update won't be triggered if topic is not changed)
* Manually delete your conversation `userId` property 
* Send yourself an `askSubscriptionStatus` broadcast, e.g. `5547be89469c64ec7d8b518d`
* Verify your user status is set to `pending` and your conversation now contains a `userId` property

This for sure could use some unit tests but it'd be best to get to this production asap per our full list broadcast currently sending. This will ensure users receiving the broadcast have their status set to `pending` -- as currently the status of users with non-backfilled conversations is not changing. It's not a critical bug, but it's a false positive on our counts of valid `'active'` and `'less'` subscribers as we send this full list broadcast.

#### Relevant tickets
#342 
#351 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -80,7 +80,31 @@ conversationSchema.statics.getFromReq = function (req) {
  */
 conversationSchema.statics.findOneAndPopulateLastOutboundMessage = function (query, req) {
   logger.debug('Conversation.findOne', query, req);
-  return this.findOne(query).populate('lastOutboundMessage');
+
+  return this.findOne(query).populate('lastOutboundMessage')
+    .then((conversation) => {
+      if (!conversation) {
+        return Promise.resolve(conversation);
+      }
+      // Our first release of Conversations did not save userId, and we haven't yet run a bulk
+      // backfill query. If req.userId exist but conversation.userId doesn't, backfill it.
+      // TODO: Remove this logic if/when a bulk backfill is run:
+      // @see https://github.com/DoSomething/gambit-conversations/issues/342#issuecomment-398834646
+      const needsUserIdBackfill = req.userId && !conversation.userId;
+      if (!needsUserIdBackfill) {
+        return Promise.resolve(conversation);
+      }
+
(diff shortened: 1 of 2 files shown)


app/models/Conversation.js
```diff
@@ -80,7 +80,31 @@ conversationSchema.statics.getFromReq = function (req) {
...
+
+      conversation.userId = req.userId; // eslint-disable-line no-param-reassign
+      logger.debug('Backfilling Conversation.userId', {
+        userId: req.userId,
+        conversationId: conversation.id,
+      }, req);
+
+      return conversation.save()
+        .then(updatedDoc => updatedDoc.populate('lastOutboundMessage'))
+        .catch(err => Promise.reject(err));
```
rapala61: What do you think of removing this catch and re-reject line? If we are not really doing anything with we could let it be caught by the `findOneAndPopulateLastOutboundMessage` caller.

aaronschachter:
thanks for the quick review! 👀 🎖

PR #354, opened 2018-06-22
remove 30008 as an undeliverable error

rapala61:
#### What's this PR do?
It removes the `30008` error from the array of errors that set an user's `sms_status` as `undeliverable`.

We need this in place before we turn the `twilio-sms-outbound-error-relay` workers back on.

- 👀 

#### Relevant Issues
Fixes [Pivotal#158556247](https://www.pivotaltracker.com/n/projects/2092729/stories/158556247)


rapala61:
Thanks for the quick 👀 @aaronschachter !

PR #361, opened 2018-07-02
Not merged: createdUser macro template [pr]

aaronschachter:
#### What's this PR do?

Adds a `createdUser` macro that sends a new corresponding `createdUser` template, welcoming new user as a subscriber.

* Hardcodes `join` trigger in `brain/macros.rive` to eventually deprecate the Contentful `join` trigger once this goes live

#### How should this be reviewed?

* Send Gambit `subscribe` and verify reply is a `createdUser` template.

* Send Gambit `join` and verify reply is a `rivescript` template with the Boom! copy

#### Any background context you want to provide?

Once this is deployed to production live, we'll change the `join` trigger with the following steps:

* Unpublish the `join` defaultTopicTrigger (https://gambit-admin.herokuapp.com/triggers#join) in Contentful, flush the G-Campaigns redis cache

* Restart G-Conversations to fetch latest Rivescript triggers

* Send Gambit `join` and verify reply is a `createdUser` template.

#### Relevant tickets
https://www.pivotaltracker.com/story/show/158416129/comments/191210520

#### Checklist
- [x] Documentation added for new features/changed endpoints - Gambit Admin is preemptively indicating that the [JOIN message is hardcoded](https://gambit-admin.herokuapp.com/triggers)
- [x] Tests added for new features/bug fixes.
- [x] Tested on staging.

PR #362, opened 2018-07-02
Adds join trigger and updates subscriptonStatusActive template

aaronschachter:
#### What's this PR do?

Implements the idea here https://github.com/DoSomething/gambit-conversations/pull/361#issuecomment-401954870 --it could make sense to avoid creating a new `createdUser` template and instead just set the `subscriptionStatusActive` template text to be used globally now that it doesn't seem we'd be asking users to respond with subscription status for a while. Ideally each broadcast should have customizable Weekly messages, vs editing our hardcoded `subscriptionStatusActive` macro reply.

Same as #361 

#### Relevant tickets
https://www.pivotaltracker.com/story/show/158416129/comments/191210520

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/brain/macros.rive
+++ b/brain/macros.rive
@@ -6,6 +6,18 @@
 + help
 @ info
 
++ subscribe
+- subscriptionStatusActive
+
+// Join currently gets overridden in Contentful: once it's unpublished, this trigger will catch.
+// @see https://app.contentful.com/spaces/owik07lyerdj/entries/2OSVbOpFGEyGOYqqcgsyqM
++ join
+@ subscribe
+
++ menu
+- menu
+
+
 + menu
 - menu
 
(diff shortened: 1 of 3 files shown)


aaronschachter:
We are still seeing a small number of `askSubscriptionStatus` responses still, so it's possible a Weekly response would send each message with "Hi it's Freddie!" per below which isn't ideal.
```
You> status

Bot> Hi it's Freddie at DoSomething! I send texts w updates on news & easy ways to take action. Want texts: A)Weekly B)Monthly C)I need more info D)Text STOP to stop 

You> weekly

Bot> Hi I'm Freddie from DoSomething.org! Welcome to my weekly updates (up to 8msg/week). Things to know: Msg&DataRatesApply. Text HELP for help, text STOP to stop.
```

I think ideally for future `askSubscriptionStatus` broadcasts, we'd be able to avoid hardcoding these questionnare replies but I'm not sure when the next time to send one out is.. so there's still a good argument for keeping #361 as a separate template in case it had to be any time soonish (cc @ngjo)

PR #363, opened 2018-07-03
Throw error if Rivescript replies are not sorted

aaronschachter:
#### What's this PR do?

Fixes up `lib/rivescript` to throw an error if  a member message is received before `sortReplies()` function has finished on app start. This way, Blink will retry the message upon errors and the user won't receive a rather unhelpful ["ERR: Replies Not Sorted" reply.](https://github.com/aichaos/rivescript-js/blob/v1.17.2/src/brain.coffee#L265)

It's not pretty but the easiest way to replicate the error is delaying setting the `ready` property in `loadingDone(batchNumber)`, e.g.:
```
function loadingDone(batchNumber) {
  logger.debug('rivescript.loadingDone', { batchNumber });

  brain.sortReplies();
  setTimeout(() => brain.ready = true, 5000);
}
```

Another cleanup TODO to avoid long loading times is that we don't need to fetch all topics to write the `menu` topic children, we can just use topics that have `defaultTopicTrigger` entries referencing them.

#### Relevant tickets
Fixes https://gambit-admin-staging.herokuapp.com/requests/9685d106-8a1c-445e-a388-6eb2a19d6f29
screen shot 2018-07-03 at 11 02 04 am

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/rivescript.js
+++ b/lib/rivescript.js
@@ -70,6 +70,9 @@ module.exports.getReply = function (conversation, inboundMessageText) {
 
   return bot.replyAsync(conversationId, inboundMessageText)
     .then((text) => {
+      if (!bot.ready) {
+        throw new Error('Still sorting Rivescript replies.');
+      }
       const res = { text };
       // The inbound message may have triggered a topic change, check bot.uservars to return topic.
       const updatedUservars = bot.getUservars(conversationId);
@@ -77,6 +80,5 @@ module.exports.getReply = function (conversation, inboundMessageText) {
       res.match = updatedUservars.__initialmatch__; // eslint-disable-line no-underscore-dangle
 
       return res;
-    })
-    .catch(err => err);
+    });
 };
--- a/lib/rivescript.js
+++ b/lib/rivescript.js
@@ -5,12 +5,17 @@ const logger = require('heroku-logger');
 const config = require('../config/lib/rivescript');
 
 let brain;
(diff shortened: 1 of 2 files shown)


lib/rivescript.js
```diff
@@ -70,13 +70,15 @@ module.exports.getReply = function (conversation, inboundMessageText) {
 
   return bot.replyAsync(conversationId, inboundMessageText)
     .then((text) => {
+      if (!bot.ready) {
```
rapala61: I'm curious, Is this a property we are monkey patching on the brain or is this prop a documented hook to check readiness in rivescript? The minor suggestion here would be a little more comments around this line to add context.

lib/rivescript.js
```diff
@@ -70,13 +70,15 @@ module.exports.getReply = function (conversation, inboundMessageText) {
...
+      }
       const res = { text };
       // The inbound message may have triggered a topic change, check bot.uservars to return topic.
       const updatedUservars = bot.getUservars(conversationId);
       res.topic = updatedUservars.topic;
       res.match = updatedUservars.__initialmatch__; // eslint-disable-line no-underscore-dangle
 
       return res;
-    })
-    .catch(err => err);
```
rapala61: good find 🕵️

lib/rivescript.js
rapala61: I have no major preference for either or. A good compromise could be to change the `bot.ready` to `bot. hasSortedReplies` + the extra comments.

PR #375, opened 2018-08-01
Refactors loading Rivescript and parsing yes/no [pr]

aaronschachter:
#### What's this PR do?

This PR refactors the `menu` Rivescript topic as `ask_yes_no`, and renames the `confirmedTopic` and `declinedTopic` macros as `saidYes` and `saidNo`, and makes the following changes as a result:

* When loading Rivescript from gambit-campaigns on app start, we no longer create Rivescript topics that include the `menu` topic for each topic we want to ask if user wants to join (e.g. `askSignup` broadcast, MENU command)

    * Instead of writing the fetched Rivescript to file, loads it via the rivescript-js `stream` function - removes `brain/contentApi.rive` from `.gitignore`

* Adds a `helpers.request.parseAskYesNoResponse` that gets the bot reply for a message within the `ask_yes_no` topic, and calls this from the catchAll macro middleware to determine if user said yes or no if the last outbound message was a yes/no question

   * Updates the inbound message macro with `saidYes` or `saidNo` to salvage broadcast stats, as the `rivescript-reply-get` middleware is now setting the macro as `catchAll` now that our campaign topics no longer include the `menu` topic.

* Verify responses to askSignup/askContinue templates work as expected (we're still parsing yes/no correctly)
* Verify Rivescript from gambit-campaigns is loaded correctly upon app start (e.g. "who is this", "hi", "mirror" are all triggers defined in Contentful)

* I've changed the `askSubscriptionStatus` topic to `ask_subscription_status` as the Rivescript parser throws an error if a topic name contains a capital letter. This error doesn't break anything as is, but ideally we should aim to update our rivescript-js to v2 and expect a Promise from the `brain.loadDirectory` call. The trouble there is that the bot will fail to load 

* This work is anticipation of supporting an `askChangeTopicBroadcast`, where we don't want to require a topic to require triggers (to include the `menu` topic) in order to parse yes/no responses to the broadcast

#### Relevant tickets
Starts on https://www.pivotaltracker.com/story/show/159413638, blocker for `askChangeTopicBroadcast` in https://www.pivotaltracker.com/story/show/157369418 

#### Checklist
- [ ] Tested on staging.


aaronschachter:
Tested on staging with an `askSignup` broadcast and verified expected behavior (and macro name change):
screen shot 2018-08-01 at 10 19 09 am

lib/helpers/request.js
```diff
@@ -170,6 +170,32 @@ function isTopicChange(req) {
...
+function parseAskYesNoResponse(req) {
+  // This gets called within the catchAll after we've already created an inbound message.
+  return helpers.rivescript.parseAskYesNoResponse(req.inboundMessage.text)
+    .then((macroName) => {
+      if (helpers.macro.isSaidYes(macroName) || helpers.macro.isSaidNo(macroName)) {
+        module.exports.setMacro(req, macroName);
+        // Overwrite the current 'catchAll' value as our saidYes or saidNo macro.
+        return req.inboundMessage.updateMacro(macroName);
+      }
+      return Promise.resolve();
```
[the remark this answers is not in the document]
aaronschachter: Removing this line results in a linting `onsistent-return ` error -- `Expected to return a value at the end of arrow function`.

server.js
```diff
@@ -11,23 +11,12 @@ require('./config/mongoose')(config.dbUri);
...
-  .then((opts) => {
-    logger.info('fetchAndWriteRivescript success', { opts });
-    rivescript.getBot();
-  })
-  // TODO: If fetchAndWriteRivescript fails, we need to retry it, not carry on as normal, as all
-  // member messages will return errors because the Rivescript bot has not finished sorting replies.
-  // @see lib/rivescript
-  .catch(error => logger.error('fetchAndWriteRivescript', { error }));
+helpers.rivescript.loadBot()
+  // TODO: Retry loading Bot if an error occurred.
```
aaronschachter: Refs #376

lib/helpers/request.js
```diff
@@ -74,12 +74,12 @@ function getCampaignActivityPayloadFromReq(req) {
 }
 
 /**
- * @param {Object}
+ * @param {Object} req
+ * @param {String} topicId
```
rapala61: Leftover comment? `topicId`.

lib/helpers/request.js
```diff
@@ -170,6 +170,32 @@ function isTopicChange(req) {
   return req.conversation.topic !== req.rivescriptReplyTopic;
 }
 
+/**
+ * @param {Object} req
+ * @return {Promise}
+ */
+function parseAskYesNoResponse(req) {
```
rapala61: Using `async/await` could probably make this function easier to implement. Example:
```
/**
 * @async
 * @param {Object} req
 * @return {Promise}
 */
async function parseAskYesNoResponse(req) {
  // This gets called within the catchAll after we've already created an inbound message.
  const macroName = await helpers.rivescript.parseAskYesNoResponse(req.inboundMessage.text);

  if (helpers.macro.isSaidYes(macroName) || helpers.macro.isSaidNo(macroName)) {
    module.exports.setMacro(req, macroName);
    // Overwrite the current 'catchAll' value as our saidYes or saidNo macro.
    await req.inboundMessage.updateMacro(macroName);
  }
}
```

lib/middleware/messages/member/macro-catch-all.js
```diff
@@ -16,32 +16,27 @@ module.exports = function catchAllReply() {
       return helpers.replies.campaignClosed(req, res);
     }
 
-    const isLastOutboundAskContinue = helpers.request.isLastOutboundAskContinue(req);
-    const isLastOutboundAskSignup = helpers.request.isLastOutboundAskSignup(req);
+    const askedContinue = helpers.request.isLastOutboundAskContinue(req);
+    const askedSignup = helpers.request.isLastOutboundAskSignup(req);
+    if (askedSignup || askedContinue) {
+      await helpers.request.parseAskYesNoResponse(req);
```
rapala61: I suggest adding a `//TODO` here to come back and wrap this in a `try/catch` block just in case we have trouble saving a user's response to the DB.

PR #376, opened 2018-08-01
Enforces HTTPS in production [pr]

rapala61:
#### What's this PR do?
It adds a middleware that will redirect any GET and HEAD requests to `https://`. Other requests like POSTs will return a `403`. Only while in the **production** environment. Local development will not be impacted.

NOTES:
There is a chance that this update will break other apps that are sending requests to G-Conversations not using the HTTPS protocol. ~~If that is the case a rollback in Staging would be in order and the update of those services would have to happen first.~~ I have to check if it's just a matter of updating the ENV variable in that service first!

- 👀 
- Run all tests

#### Relevant tickets
G-Conversations item in [Pivotal#159114978](https://www.pivotaltracker.com/story/show/159114978)

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/server.js
+++ b/server.js
@@ -7,14 +7,33 @@ require('dotenv').config();
 require('newrelic');
 
 const config = require('./config');
-require('./config/mongoose')(config.dbUri);
-
-const app = require('./app');
-const mongoose = require('mongoose');
-const logger = require('heroku-logger');
-
+const logger = require('./lib/logger');
 const helpers = require('./lib/helpers');
 const rivescript = require('./lib/rivescript');
+const app = require('./app');
+
+// Start mongoose connection
+require('./config/mongoose')(config.dbUri);
+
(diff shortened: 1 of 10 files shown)


server.js
```diff
@@ -7,14 +7,33 @@ require('dotenv').config();
...
+ */
+const db = mongoose.connection;
 
-const helpers = require('./lib/helpers');
-const rivescript = require('./lib/rivescript');
+// Register connection error listener
+db.on('error', (error) => {
+  /**
+   * TODO: This is not good, we should kill the app! If we had clustering, or a process management
+   * setup, the child process would respawn and try to re-connect.
```
aaronschachter: [Great minds!](https://github.com/DoSomething/gambit-conversations/pull/375/files#r207092172)

PR #373, opened 2018-07-25
Adds autoReplyBroadcast support [pr]

aaronschachter:
#### What's this PR do?
Makes the changes outlined in https://github.com/DoSomething/gambit-campaigns/pull/1063#issuecomment-407923816 to support autoReplyBroadcasts.

#### How should this be reviewed?
Testing Conversations against a Campaigns branch running  https://github.com/DoSomething/gambit-campaigns/pull/1063 : 

* Verify that the test `autoReplyBroadcast` 4nwTwvXmfuuYAGYgusGyyW works as expected -- the broadcast message is sent, and all subsequent replies are the `autoReply` template.

* Verify all other existing types of broadcasts work as expected (`survey_response` should still work, as we're not removing any functionality, or that hardcoded Rivescript topic -- yet)

#### Any background context you want to provide?
* We'll need to deploy https://github.com/DoSomething/gambit-campaigns/pull/1063 to production first before we can send `autoReplyBroadcast` messages on production.

* Also skips tests for now as priority is determining level of effort for handling yes/no responses with a new broadcast content type (more to come on that in a future PR)

#### Relevant tickets
https://www.pivotaltracker.com/story/show/157369418

#### Checklist
- [x] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love.

PR #377, opened 2018-08-04
Macro middleware TLC [pr]

aaronschachter:
#### What's this PR do?

This PR started out aiming to fix a bug in an autoReplyBroadcast #373 -- while in an `autoReplyBroadcast` topic, if a user sends a message that matches a default topic trigger, their topic is changed to `random` instead of remaining in the autoReplyBroadcast, as is the intent of the autoReply.

We needed an extra check in the `lib/messages/middleware/macro-parse`, but it took a bit of 🤔 to remember why there were separate Conversation `setTopic` and `changeTopic` instance methods, and it's because `setTopic` expects a string (Rivescript topic name) and `changeTopic` expects an object (returned from the [gambit-campaigns topics API](https://github.com/DoSomething/gambit-campaigns/blob/master/documentation/endpoints/topics.md#topics)). It all came back to me about how this was a bit rushed to get the Ask Subscription Status broadcast #347 out the door, and felt like time for some TLC split out the `macro-parse` middleware into separate middleware functions, and add a whole lot more unit tests and comments to better document the logic for determining how to respond to a member message.

Changes:

* Splits `macro-parse` middleware into smaller `rivescript-reply-parse`, `macro-change-topic` `macro-reply` middleware functions

* Adds more topic helper functions to deprecate `config/app/models/conversation` and DRY defining Rivescript topics via config

* Moves checks for updating user's subscription status to `pending` and avoiding unnecessary db writes out of `Conversation.setTopic`  and into `helpers.request.changeTopic` 

* Updates all middleware to call `helpers.request.changeTopic` (instead of a `req.conversation` instance method like `setTopic`) if a topic change is required

👀 

Verify expected behavior per topic changes:
* [x] Sending a topic trigger changes topic (e.g. send `mirror`, verify topic is set to Mirror Messages)
* [x] Changing topic to `ask_subscription_status` sets user `sms_status` to pending, responding with A B will update topic to random. The reply to a subsequent user message should either ask to continue the previous campaign topic, or send the `noCampaign` template if the conversation doesn't have a `campaignId`, or has `campaignId` without any active topics (e.g. send `status`. Note - previous ask subscription status broadcasts like [lVK2iLjEjIIWUSiWaSc6Y](https://gambit-admin.herokuapp.com/broadcasts/lVK2iLjEjIIWUSiWaSc6Y) won't work anymore because we changed the topic name from `askSubscriptionStatus` to `ask_subscription_status` in #375 )
* [x] Sending a legacy broadcast with a Rivescript topic set should set the topic to the Rivescript topic (e.g. broadcast [2lsqawuQSsskIWgGky6uw6](https://gambit-admin.herokuapp.com/broadcasts/2lsqawuQSsskIWgGky6uw6))
* [x] Sending a legacy broadcast with a campaign and template set should set the topic to the campaign's active topic (e.g. broadcast [1t9GKlWxAEaOgI0wysQeIM](https://gambit-admin.herokuapp.com/broadcasts/1t9GKlWxAEaOgI0wysQeIM))
* [x] Verify initial bug that started this description and rabbit hole is fixed: verify that sending a message that matches a default topic trigger while `autoReplyBroadcast` does not change the topic to `random` (e.g. broadcast [4nwTwvXmfuuYAGYgusGyyW](https://gambit-admin.herokuapp.com/broadcasts/4nwTwvXmfuuYAGYgusGyyW))
* [x] Verify signup messages work as expected (execute a `POST /messages?origin=signup` with a valid `campaignId`, e.g. 7951)

Next up for a future TLC PR is splitting out `brain/topics.rive` into separate files in a `brain/topics` dir to improve readability. It'd also make sense to create a new type of `askSubscriptionStatus` broadcast to avoid hardcoding replies for that topic, and instead using the topic as a helper for parsing per our `ask_yes_no` in #375 

#### Relevant tickets

#### Checklist
- [x] Tested on staging.


lib/helpers/request.js
```diff
@@ -206,6 +215,23 @@ function setCampaign(req, campaign) {
   }
 }
 
+/**
+ * @param {Object} req
+ * @param {Conversation} conversation
+ */
+function setConversation(req, conversation) {
+  req.conversation = conversation;
+  req.currentTopicId = conversation.topic;
```
rapala61: One observation. I've noticed a few `setX` function setting more than what their name suggest (side-effects). Ideally, all the `setX` functions should be very atomic and do 1 thing very well. If we needed a function that does a few things, then we can try naming it in a way that makes it obvious that multiple things are being done and call the side-effects from there. If that is not entirely possible (naming), we can use comments aid in clarifying the purpose of the function, like:

```
// sets conversation, currentTopic, and lastOutboundMessage in the req object
function setConversationReqParams(req, conversation) {
  module.exports.setConversation(req, conversation);
  module.exports.setCurrentTopic(req, conversation.topic);
  module.exports.setLastOutboundMessage(req, conversation.lastOutboundMessage);
}
```

I've noticed this here too: 
- [`setTopic`](https://github.com/DoSomething/gambit-conversations/pull/377/files#diff-8d18d3bce63d43406f0d205e2c99b48dR257)
- [`changeTopic`](https://github.com/DoSomething/gambit-conversations/pull/377/files#diff-8d18d3bce63d43406f0d205e2c99b48dR16)

aaronschachter: Definitely a valid observation, and would be great to cleanup in a future PR - I agree the names can be misleading.

- `setTopic` in particular could be cleaned up by just referencing a `req.topic.campaign` instead of `req.campaign` whenever we need it in various middlewares

- `changeTopic` - I've been wondering about a like `updateTopicIfShouldTopicChange`.. in particular for `setConversation`, we save the `campaignId` on the conversation simply to find a campaign topic if the conversation gets updated to a non campaign topic -- for the sake of when we're out of checks to make in the catchall, let's prompt to continue the last active campaign the user participated. I could see this potentially getting phased out by all topics having an autoReply.. and no longer needing to save a `campaignId` property on a conversation.

aaronschachter:
Thanks so much for the review @rapala61 -- sorry about the size of this PR, it got a bit 🐘 adding all those tests..but also wanted to make sure I had a test in for the initial bug.

PR #380, opened 2018-08-12
Post campaign activity if autoReply topic has campaign [pr]

aaronschachter:
#### What's this PR do?
If a user sends a message while in an `autoReply` topic that has a campaign set, post it to the gambit-campaigns `campaignActivity` endpoint to create a signup. This starts on deprecating the need for the `externalPostConfig` (but isn't ready to deprecate until we add entry/welcome messages for web signups or keywords that would change the conversation topic to the autoReply.

On staging, broadcast 6zj3wladnGyWuMWAUYyg8c will change a user's topic to an autoReply topic with a campaign set (2Wzzquygx2wwMWe8kQAMgc). Verify upon entering the topic, a signup is posted for the campaign (2299)

#### Relevant tickets

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -5,7 +5,6 @@ const { STATUS_CODES } = require('http');
 const logger = require('./logger');
 const config = require('../config/lib/helpers');
 const helpers = require('./helpers/index');
-const InternalServerError = require('../app/exceptions/InternalServerError');
 
 // TODO: Move contents of this helper.js file into lib/helpers/index.js or to modular helpers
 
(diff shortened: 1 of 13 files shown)


lib/helpers.js
```diff
@@ -14,6 +14,22 @@ Object.keys(helpers).forEach((helperName) => {
   module.exports[helperName] = helpers[helperName];
 });
 
+/**
+ * @param {Object|Error}
+ * @return {String}
+ */
+module.exports.parseMessageFromError = function (error) {
+  // Check for a response body with an error message.
+  if (error.response && error.response.body && error.response.body.error) {
```
rapala61: Small suggestion, we can take the `error.response && error.response.body` into it's own boolean variable to DRY the second `if statement`:
```
const hasResponseBody = error && error.response && error.response.body;

if (hasResponseBody && error.response.body.error) { ... }

if (hasResponseBody) { ... }
```

lib/helpers.js
```diff
@@ -14,6 +14,22 @@ Object.keys(helpers).forEach((helperName) => {
...
+    return error.response.body.error.message;
+  }
+  // Check for a response body with a message.
+  if (error.response && error.response.body) {
+    return error.response.body.message;
+  }
+  return error.message ? error.message : error.toString();
```
rapala61: If the [pr] suggestion is implemented then this wouldn't be a problem. Otherwise, if `error` = `undefined` then we would have an `Uncaught TypeError: Cannot read property 'message' of undefined` exception.

Issue #170, opened 2017-10-17
Include Source Detail in Northstar + Rogue requests


aaronschachter:
Both Northstar and Rogue accept a `source_detail` parameter we're not sending anything for.

* [ ] User
* Sounds like we should be sending a keyword, if sent? https://dosomething.slack.com/archives/C2C8NLNAY/p1507925259000068

* [ ] Signup
* [ ] Post:

* Ideas for sending various details: `keyword-mirror`, `keyword-menu`, `broadcast-59cad9d9b76942000496f4df`. If we went this route, it'd make sense to send a detail with `keyword-` prefix to our User source detail

cc @jamjensen @justkika @sheyd

sheyd:
@aaronschachter I like the initial thoughts. I think it's at least parity with what we have now from MoCo. I'll let @jamjensen add if he has any thoughts on if more details are required, or has any syntactical preferences. Otherwise, I think 👍  on this and it addresses our current needs.

aaronschachter:
@jamjensen @sheyd -- Confirming we're 👍  to change the prefix a Northstar User's `source_detail` with the `keyword-`? e.g. `keyword-mirror` instead of `mirror`

jamjensen:
@aaronschachter yes, this looks good to me. Will `source-detail` be limited to keyword info, or will it also include, `broadcast-number`?

DFurnes:
It might be nice to mirror the source details we've been [saving from Mobile Commons](https://github.com/DoSomething/northstar-mobilecommonsd/blob/a56b366339713d370dc13948f2fe9deb87d9559f/app/Jobs/SendUserToNorthstar.php#L98-L112), so `keyword/mirror` or `broadcast/1234`. (Especially if we decide to standardize all SMS sources as a single `sms` in the future!)

aaronschachter:
That's the ticket @DFurnes 💸  

If there's a Broadcast ID set in the request, we can pass it for User source @jamjensen. In theory, I don't think we'd ever create a User via Gambit when a member is responding to a broadcast -- they should be in Customer.io / Northstar already?

jamjensen:
@aaronschachter agreed, that member should be in C.io / Northstar already.

I was referring to the `source_detail` as it relates to the signup. If Gambit is posting the signup to Rogue, can it also pass along `source_detail` if the signup comes from a broadcast or keyword?

aaronschachter:
[...]

PR #385, opened 2018-08-20
Fixes outbound broadcast message count

### LINKED ISSUE #384: Outbound replies in an AskYesNo broadcast topic should not contain a broadcastId
_BUG_
-------------------
#### Current Behavior
- A `GET /broadcasts/:id` request was breaking for an askYesNo broadcast, because the aggregator doesn't expect to find any outbound broadcast messages with macros (hence the hotfix in #383). This occurs because the askYesNo's reply messages are being saved with a `broadcastId`, giving us an inaccurate count of outbound broadcast messages (we should only save an outbound `broadcastId` for the broadcast message sent via a `POST /messages?origin=broadcast` request, excluding our outbound replies sent after the broadcast)

#### Desired Behavior
- Stats for Ask Yes No's should refer to the total count of outbound broadcast messages created from `POST /messages?origin=broadcast` requests sent 

#### Steps to Replicate
- Send yourself an askYesNo broadcast, e.g. 4yFELpwjwA6EMaeYggu4qU and respond with a yes. Verify stats indicate 2 outbound messages were sent, instead of the expected 1.

#### Why This Matters

We use these stats to 👀 that the total number of outbound messages correponds to the webhooks created from a Customer.io triggered campaign. Including the total # of replies in our count of outbound leads to false alarms. Ideally these stats will be deprecated by Looker some time in the not-too-distant future.

https://dosomething.slack.com/archives/C2C8NLNAY/p1534792849000100

aaronschachter:
#### What's this PR do?

Refactors `helpers.request.executeSaidYesMacro` to avoid relying on a `req.broadcastId` to include a `broadcastId` parameter for a gambit-campaigns `POST /campaignActivity` request. Also removes the broadcast stats caching in a `GET /broadcasts/:id` request, as this isn't utilized.

* Instead of removing the cache helper completely now that the last remaining `broadcastStats` cache is gone, I've renamed it as a `rivescript` cache in anticipation of using Redis to have a single source of Content API Rivescript that all dynos can share, instead of each dyno needing to query gambit-campaigns on start to load the Rivescript into memory.

Send yourself an askYesNo broadcast, e.g. 4yFELpwjwA6EMaeYggu4qU and respond with a yes. Verify stats indicate only 1 outbound message was sent (not 2 per #384).

#### Relevant tickets
Fixes #384 

#### Checklist
- [x] Tested on staging.


lib/helpers/request.js
```diff
@@ -124,7 +126,6 @@ function getCampaignActivityPayloadFromReq(req) {
     postType: req.topic.postType,
     text: req.inboundMessageText,
     mediaUrl: req.mediaUrl,
-    broadcastId: req.broadcastId,
```
rapala61: Flagging that this change might have a bigger side effect than intended since it will stop all calls to [`postCampaignActivityFromReq `](https://github.com/DoSomething/gambit-conversations/blob/master/lib/helpers/request.js#L243) from including the `broadcastId`.

aaronschachter: That's a valid flag... we don't expose the `broadcastId` value saved on our `signup` collection anywhere in API but there's always been talk of using it to backfill the `source_detail` of a signup. A `textPostBroadcast` is an example of where both a signup and post would be created with a `broadcast/:broadcastId` detail (refs #170), and an instance where this change is not ideal.

One idea that comes to mind is modifying `helpers.request.postCampaignActivityFromReq` to accept an additional `broadcastId` parameter, and modify this line to set req.broadcastId equal to `broadcastId || req.broadcastId`. It may make sense to rename the functions `postCampaignActivityFromReq` and `getCampaignActivityPayloadFromReq` to just `postCampaignActivity` and `getCampaignActivity` per https://github.com/DoSomething/gambit-campaigns/pull/1075#discussion_r210470772

aaronschachter: I'll make this change in a next PR, as it'll be a bit more renaming.

PR #389, opened 2018-08-22
Deprecates menu macro and askSignup replies

aaronschachter:
#### What's this PR do?

Some post legacy broadcast housekeeping:

* Removes the `menu` macro and middleware per https://www.pivotaltracker.com/story/show/159796050. Reasons why:

   * it's not business critical, rarely used or referred to 

   * relies on a nested topics array in a `GET /campaigns/:id`  response -- and we're looking to deprecate the list of nested topics to simplify that endpoint

    * Editors don't ever set overrides for the `askSignup` templates, so the menu itself isn't ever all that useful to know what you're signing up for

*  Now that the `askSignup` legacy broadcast has been deprecated in #387 and we've removed the `menu` command, there aren't any other entry points that would trigger the `askSignup` reply and the followup `confirmedSignup`, `declinedSignup`, and `invalidAskSignupResponse` replies -- they've been deprecated by the new `askYesNo`, `invalidAskYesNoResponse`, `saidNo`, and `saidYes` replies

* Verify a MENU command no longer sends an `askSignup` for a random campaign

* Verify an `askYesNo` broadcast works as expected, e.g. 4K6E6y2wnY6aesy6KgOQoO

* While in a topic (e.g Pump it up per broadcast above ☝️) trigger a quick reply like "Hello". Verify that sending a subsequent `catchAll` message while send the `askContinue` template, prompting to continue the last campaign. 

“10. Sometimes removing a feature is just as valuable as building one.” — `@bephf `https://medium.com/making-donorschoose-org/product-principles-8355e4753eb7

Will need to update the hardcoded `declinedContinue` template values in the Content API to no longer prompt user to text MENU to find a different action.

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/159796050

#### Checklist
- [x] Tested on staging.


aaronschachter:
Thanks for reviewing! Definitely a great post, @mshmsh5000 shared it in the [#team-product channel last week.](https://dosomething.slack.com/archives/C09ANFQLA/p1534889730000100)

PR #390, opened 2018-08-23
Rivescript cache part 1 [pr]

aaronschachter:
#### What's this PR do?

Fills out the rivescript cache helper introduced in #385 to regularly reload the Rivescript bot replies. 

* Modifies the `getBotReply` rivescript helper to reload the bot if the Rivescript cache is empty, deprecating the need to load the bot in `server.js` 

Easiest way to verify expected behavior is to set your local `DS_GAMBIT_CONVERSATIONS_RIVESCRIPT_CACHE_TTL` config setting to a small value, and verify that default topic triggers are fetched and the bot is reloaded once the cache expires.

Part 2: Provide an endpoint to clear the Rivescript cache, so editors can manually clear cache to verify triggers

Part 3: Which truly allows editors to clear cache, introduce Redis caching so all dynos load the additional rivescript from a single source. When this is introduced, we'll no longer need to serve the `GET /defaultTopicTriggers` as a cached list of default topic triggers, but instead can query Contentful without caching all results.

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/158037562 at a bare minimum

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/rivescript.js
+++ b/lib/rivescript.js
@@ -29,17 +29,13 @@ function loadingDone() {
  *
  * @param {Array} rivescripts
  */
-function createNewBot(rivescripts) {
+function loadBotWithRivescripts(rivescripts = []) {
+  logger.debug('loadBotWithRivescripts');
+  additionalRivescripts = rivescripts;
   try {
-    additionalRivescripts = rivescripts;
-    brain = new RiveScript({
-      debug: config.debug,
-      concat: config.concat,
-    });
-    // Note: The v2 release of rivescript-js (still in alpha) returns a Promise, which means any
-    // Rivescript errors found when loading will stop the bot from loading entirely. We don't have
-    // a need to pass a Contentful ID to a reply because we're no longer adding specific Rivescript
-    // for our Content API topics -- but holding off on updating the package for now.
+    if (!brain) {
+      brain = new RiveScript({ debug: config.debug, concat: config.concat });
+    }
     brain.loadDirectory(config.directory, loadingDone, handleError);
   } catch (err) {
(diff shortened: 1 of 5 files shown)


lib/rivescript.js
```diff
@@ -29,17 +29,13 @@ function loadingDone() {
  *
  * @param {Array} rivescripts
  */
-function createNewBot(rivescripts) {
+function loadBotWithRivescripts(rivescripts = []) {
```
rapala61: Question, why don't we set `hasSortedReplies` to `false` at some point when reloading the bot w/ new rivescript?

aaronschachter: That's a very good question, I think I missed adding that in. Will update

rapala61:
This Part 1 seems good to me. Part 3 sounds to me like should come before Part 2, since we would need to have the Redis cache already working before we could invalidate it via a request to a single endpoint.

One potential bug that comes to mind is that even though each Dyno will be able to fetch the cache from a single source (Redis), each Dyno still has to go through checking if the cache exists individually before re-fetching. This can result in a condition where only a few Dynos found out about a missing cache and re-fetch, where the others never perceive a change (the cache is there when they receive the new traffic, hence wouldn't call `loadBot`).

I think is worth thinking about that scenario and how likely it is to happen. A solution that involves some sort of Rivescript cache manager that orchestrates a downstream event to all Dynos to re-fetch comes to mind.

rapala61:
Another idea that comes to mind involves calculating a hash from the fetched rivescript contents and storing that in the cache key. We can talk through this if it sounds like a reasonable approach to you!

Issue #393, opened 2018-08-30
Fetching and sorting Rivescript on demand


aaronschachter:
_FEATURE OVERVIEW_
-------------------
#### User Story

As a developer, I want to enable editors to force all Conversations dynos to reload with the latest Rivescript, avoiding the need for each dyno to regularly query the Content API once its local cache clears.

Refs https://www.pivotaltracker.com/story/show/158037562

#### Additional Information

Options we discussed today, potentially in order of increasing T-Shirt size:

* MVP? Using redis per #390 -- cache the parsed Rivescript in Redis. Allow editors to manually clear the cache via query string parameter, which they can trigger via Gambit Admin

* A pub sub implementation, where each dyno listens for an event that should trigger a bot reload

 * A microservice or Serverless function that exists to fetch Rivescript from the Content API, and provide an endpoint that Conversations can post to, using the response to determine the outbound reply for an inbound message 

#### Why This Matters

Editors aren't easily able to test their content without a dev going through manual process described above.

PR #392, opened 2018-08-29
Adds a GET /rivescript endpoint

aaronschachter:
#### What's this PR do?

Adds a `GET /rivescript` endpoint that returns [deparsed Rivescript](https://github.com/aichaos/rivescript-js/blob/master/docs/rivescript.md#data-deparse). Also updates rivescript-js to v2 to take advantage of the async [`loadDirectory`.](https://github.com/aichaos/rivescript-js/blob/master/docs/rivescript.md#async-loaddirectory-string-path)

#### How should this be reviewed?
Execute a `GET /rivescript` request, verifying that it loads the bot when Rivescript cache is not set. 

#### Any background context you want to provide?

Per part 2 in #390 - this endpoint will eventually support a query parameter to clear Redis cache once we're using Redis to trigger fetching and sorting the Rivescript triggers defined in the Content API.

Skipping on some unit tests for now to try to get a step 3 proof of concept with Redis in a new branch.

#### Relevant tickets
https://www.pivotaltracker.com/story/show/158037562

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tests added for new features/bug fixes.
- [ ] Tested on staging.

PR #394, opened 2018-08-30
Adds Redis to rivescript cache, clear cache parameter [pr]

aaronschachter:
#### What's this PR do?

* Refactors the rivescript cache to use Redis, following by @rapala61's example in gambit-campaigns (e.g. https://github.com/DoSomething/gambit-campaigns/pull/995)

* Adds a `cache` query parameter to the `GET /rivescript` endpoint introduced in #393, unblocking adding a cache clear button into Gambit Admin to trigger dynos to stream Rivescript from our cache, and call `sortReplies`
* I've added the Heroku Redis to staging to test this PR; we will need to add to prod if we go this route.

Verify a `GET /api/v2/rivescript?cache=false` request queries the Content API, verify excluding the query parameter does not query the Content API if the bot has already been loaded.

* Our cache clearing request needs to reload the bot, which is why it makes sense to set the `hasSortedReplies` to false per https://github.com/DoSomething/gambit-conversations/pull/390#pullrequestreview-150584626 -- but I think we may possibly still be in the same boat of each inbound message being retried multiple times until a reply is sent (we've seen a surge of the "Still sorting bot replies, please retry your message" errors since deploying [3.1.0](https://github.com/DoSomething/gambit-conversations/releases/tag/3.1.0) today).

* Similar to #392, this skips out on unit tests at the moment for sake of validating whether this approah will work for us.

#### Relevant tickets
Potentially fixes https://www.pivotaltracker.com/story/show/158037562 and #393 

#### Checklist
- [x] Tested on staging.


lib/helpers/cache.js
```diff
@@ -1,19 +1,25 @@
...
+const redisClient = require('../../config/redis')();
 
 const config = require('../../config/lib/helpers/cache');
 
-// TODO: Add Redis
-const rivescriptCache = new Cacheman(config.rivescript.name, config.rivescript.ttl);
+const redisEngine = new RedisEngine(redisClient);
+
+const rivescriptCache = new Cacheman(config.rivescript.name, {
+  ttl: config.rivescript.ttl,
```
rapala61: We can't ditch the `ttl` in cacheman so we would have to set it to something ridiculous like [100 years](https://github.com/cayasso/cacheman/issues/31#issuecomment-247823203).

lib/rivescript.js
```diff
@@ -14,18 +14,22 @@ function getBot() {
...
-
+/**
+ * @param {Array} rivescripts
+ */
 function streamAndSortRepliesWithRivescripts(rivescripts) {
-  rivescripts.forEach(item => brain.stream(item, logStreamError));
+  rivescripts.forEach(item => brain
+    .stream(item, error => logger.error('error streaming rivescript', { error })));
   logger.info('rivescript.brain.stream success');
   brain.sortReplies();
```
rapala61: Just double checking, is `sortReplies` a promise? since we updated rivescript to 2.x which uses async/await.

rapala61: Great, thanks for checking!

lib/helpers/rivescript.js
```diff
@@ -8,29 +8,35 @@ const config = require('../../config/lib/helpers/rivescript');
...
 
 /**
- * @param {Object} query
  * @return {Promise}
  */
 async function getRivescripts() {
   logger.debug('getRivescripts');
   const cache = await helpers.cache.rivescript.get(cacheKey);
   if (cache) {
+    logger.debug('rivescript cache hit');
```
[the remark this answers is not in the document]
aaronschachter: Yep -- we have similar debug statements in other catch getters to verify cache hits.

aaronschachter:
Update re: https://github.com/DoSomething/gambit-conversations/pull/394#issuecomment-417203250 is that this seems to be something up with our long code and AT&T, and is unrelated to this branch. See [Slack thread](https://dosomething.slack.com/archives/C2C8NLNAY/p1535674182000100).

aaronschachter:
I've added unit tests and tested the latest branch on staging, looking good. Also added the Heroku Redis add-on to production, so I think this is good to go!

PR #397, opened 2018-09-04
Fixes bug in reloading Rivescript after cache clear

aaronschachter:
#### What's this PR do?

Fixes bug where a Rivescript cache clear adds and sorts triggers from the Content API without ever purging the existing in-memory replies, creating duplicate triggers.  Found this by testing out the cache clear in https://github.com/DoSomething/gambit-admin/pull/69, where React began throwing errors for duplicate keys when rendering a `TriggerListItem` component.

Verify the number of triggers found in the `data.topics.random` array property remains the same upon testing a few `GET /rivescript?cache=false` requests.

We haven't merged #394 into production yet, we'll want to get this fix in before we do.

#### Relevant tickets
https://www.pivotaltracker.com/story/show/158037562

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/middleware/rivescript/index.js
+++ b/lib/middleware/rivescript/index.js
@@ -1,13 +1,18 @@
 'use strict';
 
 const helpers = require('../../helpers');
+const logger = require('../../logger');
 
 module.exports = function getRivescript() {
   return (req, res) => {
     const resetCache = req.query.cache === 'false';
 
     return helpers.rivescript.getDeparsedRivescript(resetCache)
-      .then(data => res.send({ data }))
+      .then((data) => {
+        const triggers = data.topics.random;
+        logger.debug('data.topics.random', { count: triggers.length });
+        return res.send({ data });
+      })
       .catch(err => helpers.sendErrorResponse(res, err));
   };
 };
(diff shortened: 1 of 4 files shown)


aaronschachter:
Deployed this to staging, and verified expected behavior by sending cache query params via https://github.com/DoSomething/gambit-admin/pull/69, which has been deployed to Gambit Admin staging.

rapala61:
Hi @aaronschachter, I don't think I understand the fix. It seems that the new code is doing the same as the old, but it's just organized better. Was the fix just to set `hasSortedReplies = false;` after creating the new bot, vs before as in the old code?

aaronschachter:
Ah, I should have spelled it out a bit better: The fix is that `loadBotWithRivescripts` calls `createNewBot` instead of `getBot`. I couldn't find a way in [rivescript-js](https://github.com/aichaos/rivescript-js/tree/master/docs) to erase the bot brain, so the `loadBotWithRivescripts` call creates new one to overwrite the `brain` const.

rapala61:
Ohh, I see what you mean now. Thanks for the spelling it out 😄.

PR #398, opened 2018-09-05
Almost fixes cache clear for GET /rivescript

aaronschachter:
#### What's this PR do?

This PR passes a `true` resetCache arg from the `helpers.rivescript.getDeparsedRivescript` to the `helpers.rivescript.loadBot`, to force fetching Rivescript from the Content API and make the cache clear parameter actually work 🤦‍♂️ 

This PR also passes a `cache` query parameter in a `GET /defaultTopicTriggers` request to the Content API, which will need to be implemented in order to truly refresh the bot with the latest default topic triggers upon a Rivescript cache clear.

I've been modifying the response of our [test "puppetsloth" trigger](https://app.contentful.com/spaces/owik07lyerdj/entries/1DiZPY8uyEWg0mSeWQkuEk) to verify expected behavior, but it'll be a bit more straight forward to test once a cache clear parameter is [introduced in Content API](https://github.com/DoSomething/gambit-campaigns/blob/master/documentation/endpoints/defaultTopicTriggers.md), as currently the steps to replicate locally are trickier when both apps share same redis instance.

Once we deploy our new release with cached Rivescript in redis, we'll eventually remove the cached list of all defaultTopicTriggers in the Content API to avoid unnecessary complexities.

#### Relevant tickets
https://www.pivotaltracker.com/story/show/158037562

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/rivescript.js
+++ b/lib/helpers/rivescript.js
@@ -43,7 +43,7 @@ async function getRivescripts(resetCache = false) {
  * @return {Promise}
  */
 function fetchRivescripts() {
-  return gambitCampaigns.fetchDefaultTopicTriggers()
+  return gambitCampaigns.fetchDefaultTopicTriggers({ cache: false, limit: 150 })
     // TODO: Check our res.meta to determine whether to fetch more triggers.
     .then((res) => {
       logger.debug('fetchDefaultTopicTriggers success', { count: res.data.length });
--- a/lib/helpers/rivescript.js
+++ b/lib/helpers/rivescript.js
@@ -14,25 +14,37 @@ const cacheKey = config.cacheKey;
  */
 async function getDeparsedRivescript(resetCache = false) {
   if (!rivescript.isReady() || resetCache === true) {
-    await module.exports.loadBot();
+    await module.exports.loadBot(resetCache);
   }
   // @see https://github.com/aichaos/rivescript-js/blob/master/docs/rivescript.md#data-deparse
   return rivescript.getBot().deparse();
 }
 
 /**
+ * @param {Boolean} resetCache
  * @return {Promise}
  */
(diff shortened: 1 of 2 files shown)


aaronschachter:
I've been testing this in tandem with https://github.com/DoSomething/gambit-campaigns/pull/1083.. and it looks like we'll definitely run into the bug described by @rapala61  in https://github.com/DoSomething/gambit-conversations/pull/390#issuecomment-416981543. The first cache clear will always result in a latest change, but pinging the `GET /rivescript` cache endpoint on staging regularly does not always return the latest change.

Per the [following comment in that PR: ](https://github.com/DoSomething/gambit-conversations/pull/390#issuecomment-416984684)
> Another idea that comes to mind involves calculating a hash from the fetched rivescript contents and storing that in the cache key

That sounds pretty reasonable as far as the smallest lift to fix this last bug 🐞 ....

aaronschachter:
A kind of hotfix that comes to mind is setting our Rivescript cache to something a whole lot smaller than a year. Once a dyno restarts, it would be refreshed, as the Rivescript bot that's created upon app start won't have sorted replies.

rapala61:
> A kind of hotfix that comes to mind is setting our Rivescript cache to something a whole lot smaller than a year. Once a dyno restarts, it would be refreshed, as the Rivescript bot that's created upon app start won't have sorted replies.

If I understand correctly, we would relay on the daily Dyno restarts to update the the Rivescript cache upon restart?

aaronschachter:
That's what I was thinking about, but I think I have a solution where we check the bot's loaded Rivescript vs what's in the cache upon each request to get a Rivescript reply. I've edited the title of this PR and going to close as is, to open a separate one to fix cache clear across all dynos.

PR #399, opened 2018-09-06
Reload bot when cached Rivescript does not match current Rivescript

aaronschachter:
#### What's this PR do?

Fixes #398 by saving our `additionalRivescript` data from the Content API to memory, in checking whether it matches the cached Rivescript whenever a `GET /rivescript` or `POST /messages?origin=member` request is received.

I've been testing by editing our dev test trigger, puppetsloth, and verifying that a `GET /rivescript?cache=false` refreshes the cache, and that subsequent requests without the cache param (or for member messages) return the updated reply copy. Will deploy to staging to verify that each subsequent `GET /rivescript` request returns the updated reply, as each dyno will check to see whether the saved `additionalRivescript` const is equal to the cached Rivescript value.

Unblocks deploy of next release with Redis caching.

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/158037562 and 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/rivescript.js
+++ b/lib/helpers/rivescript.js
@@ -160,7 +160,7 @@ function loadBot(resetCache = false) {
  * @return {Promise<boolean>}
  */
 async function isBotCurrent() {
-  if (!rivescript.ready) {
+  if (!rivescript.isReady()) {
     return false;
   }
   const cache = await helpers.cache.rivescript.get(cacheKey);
--- a/lib/helpers/rivescript.js
+++ b/lib/helpers/rivescript.js
@@ -10,19 +10,9 @@ const config = require('../../config/lib/helpers/rivescript');
 const cacheKey = config.cacheKey;
 
 /**
- * @param {Boolean} resetCache
- * @return {Promise}
+ * @return {Object}
  */
-async function getDeparsedRivescript(resetCache = false) {
-  if (resetCache === true) {
-    await module.exports.loadBot(true);
-  } else {
-    const isCurrent = await module.exports.isBotCurrent();
-    if (isCurrent === false) {
-      await module.exports.loadBot();
-    }
-  }
-
+function getDeparsedRivescript() {
   // @see https://github.com/aichaos/rivescript-js/blob/master/docs/rivescript.md#data-deparse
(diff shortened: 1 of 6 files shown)


aaronschachter:
Tested on staging and this branch is not working as expected yet, as it seems to reload the bot upon each inbound message 🙁, edited title as WIP

lib/rivescript.js
```diff
@@ -60,15 +69,18 @@ function isReady() {
  * Creates a Rivescript bot and loads it with our hardcoded rivescript as well as rivescripts passed
  * in a rivescripts array property.
  *
- * TODO: Don't provide an empty array, return a rejected error.
  * @param {Array} rivescripts - Rivescript parsed from the Content API
  */
-function loadBotWithRivescripts(rivescripts = []) {
+function loadBotWithRivescripts(rivescripts) {
```
rapala61: Since we are moving towards `async/await` could we refactor this method to use it?

lib/helpers/rivescript.js
```diff
@@ -13,9 +14,15 @@ const cacheKey = config.cacheKey;
  * @return {Promise}
  */
 async function getDeparsedRivescript(resetCache = false) {
-  if (!rivescript.isReady() || resetCache === true) {
-    await module.exports.loadBot(resetCache);
+  if (resetCache === true) {
+    await module.exports.loadBot(true);
+  } else {
+    const isCurrent = await module.exports.isBotCurrent();
+    if (isCurrent === false) {
```
rapala61: Curious why this conditional is not just `if (!isCurrent)`.

rapala61: Another observation. The `isBotCurrent` function checks if the bot has sorted replies,  as well as if cache exists. I think that makes it a little hard to understand what this function is really saying about the state of the bot and it's currentness. 

I think if we could check the state of the bot's readiness outside of the cache currentness it would make it easier. For example. Adding a check for if the bot is ready before actually calling `getDeparsedRivescript` and if not ready, then make it ready and return with a `retry your request` error to keep it consistent.

PR #404, opened 2018-09-19
Adds twilioStudio origin to POST /v2/messages

aaronschachter:
#### What's this PR do?

Adds support for a Twilio Studio prototype to get the Rivescript reply (and loaded user) for an execution triggered via incoming message. If the origin is set to `twilioStudio`, we do not post to the Twilio API to send an outbound reply; the reply is handled within the Studio flow execution steps.

Post a Twilio message payload to `POST /v2/messages?origin=twilioStudio` for your mobile number. Verify that your conversation state is not changed, and a message is not sent to your phone (instead the HTTP response data should be set an object with  `user` and  `reply` object properties.

It makes sense to casually maintain a [working prototype of Twilio Studio flow integration](https://docs.google.com/document/d/1kWvVOk00ZE99XPyuRjcW0iONUIR_oeAMf_R8zGrZnd0/edit?ts=5b9fb80b#) as new Studio features get released. Here's a snapshot of the prototype flow currently, supporting mock voting plan keyword, voting plan broadcast, auto reply broadcast, and text post broadcast conversations.

screen shot 2018-09-19 at 3 21 01 pm

#### Relevant tickets
https://www.pivotaltracker.com/story/show/158991881

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/middleware/messages/member/rivescript-reply-get.js
+++ b/lib/middleware/messages/member/rivescript-reply-get.js
@@ -15,7 +15,16 @@ module.exports = function getRivescriptReply() {
         helpers.request.setMacro(req, req.rivescriptReplyText);
       }
 
-      return next();
+      if (!helpers.request.isTwilioStudio(req)) {
+        return next();
+      }
+
+      const data = {
+        user: req.user,
+        reply: rivescriptRes,
+      };
+
+      return res.send({ data });
     })
     .catch(err => helpers.sendErrorResponse(res, err));
 };
--- a/lib/middleware/messages/member/rivescript-reply-get.js
+++ b/lib/middleware/messages/member/rivescript-reply-get.js
@@ -6,6 +6,10 @@ const helpers = require('../../../helpers');
 module.exports = function getRivescriptReply() {
   return (req, res, next) => helpers.request.getRivescriptReply(req)
     .then((rivescriptRes) => {
+      if (helpers.request.isTwilioStudio(req)) {
+        return helpers.response.sendData(res, { user: req.user, reply: rivescriptRes });
+      }
+
       req.rivescriptReplyText = rivescriptRes.text;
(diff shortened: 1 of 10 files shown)


aaronschachter:
It's not looking like Studio is a viable option in the near future because we're limited to a single Flow to handle all messaging on our shortcode. I do think it's worth merging this branch (although could be open renaming the `twilioStudio` value as something else, or as an entirely different endpoint as this `origin` type doesn't create any documents in the `messages` collection -- which is to keep things simple) so our [prototype](https://user-images.githubusercontent.com/1236811/45770498-67f18d80-bbf7-11e8-8137-52b008361716.png) can remain functional, and as something to potentially something to iterate on per potential future Studio releases.

Here's a rough sketch of how a single Flow could handle executions created by both inbound messages (e.g. our campaign signup keywords) and API requests (broadcasts, signups, etc) dynamically, but it only draws out branches for Voting Plans, Text Posts, and [DSS](https://dosomethingstrategic.org/) subscribers -- missing [split conditions](https://www.twilio.com/docs/studio/widget-library#split-based-on) for Ask Yes No, Photo Posts, Ask Subscription Status):

image from ios

lib/middleware/messages/member/rivescript-reply-get.js
```diff
@@ -15,7 +15,16 @@ module.exports = function getRivescriptReply() {
...
+      if (!helpers.request.isTwilioStudio(req)) {
+        return next();
+      }
+
+      const data = {
+        user: req.user,
+        reply: rivescriptRes,
+      };
+
+      return res.send({ data });
```
rapala61: What do you think of moving this reply to a function in `/lib/helpers.js`  or `/lib/helpers/replies.js` to be more consistent?

PR #408, opened 2018-09-29
Refactor profile updates: part 1 [pr]

aaronschachter:
#### What's this PR do?

We update a user profile every time they send a message. This PR deprecates the `lib/middleware/messages/member/user-update` function that would handle this, moving the profile update request into a new `updateByMemberMessageReq` function in `lib/helpers/user`. This PR modifies the `sendReply` function in `lib/helpers/replies` to call `updateByMemberMessageReq`:  all of the middleware functions that deliver outbound replies to users in our `app/routes/messages/member` route end up calling `sendReply` by the way of the various `helpers.replies` functions used.

This PR also refactors determining the profile fields and values to update by adding them to the config for the user helper, which aims to make it easier to add saving profile values for the new fields we need for Voting Plan conversations in https://www.pivotaltracker.com/story/show/160824069

Verify subscription status keywords like JOIN, LESS, STOP work as expected: updating your user's `sms_status` accordingly in Northstar.

Part 2 will either consolidate `lib/helpers/subscription` into `lib/helpers/user` (which is the approach I'm leaning toward), or rename as `lib/helpers/profile`. This is where we'll have macros and update definitions for upcoming like macros like `votingPlanStatusVoting`, `votingPlanStatusNotVoting`, `votingPlanMethodOfTransportWalk`, `votingPlanMethodOfTransportBike`

#### Relevant tickets
https://www.pivotaltracker.com/story/show/160824069

#### Checklist
- [x] Tested on staging.


lib/helpers/user.js
```diff
@@ -71,6 +90,7 @@ module.exports = {
     const data = {
       source: req.platform,
       mobile,
+      // TODO: Verify - I don't think we need to send a password?
```
DFurnes: Yes, definitely no need to send a password!

lib/helpers/user.js
```diff
@@ -18,6 +18,25 @@ function setPendingSubscriptionStatusForUserId(userId) {
...
 }
 
+/**
+ * @param {Object} req
+ * @return {Promise}
+ */
+function updateByMemberMessageReq(req) {
+  const payload = {};
+  try {
+    underscore.extend(payload, module.exports.getDefaultUpdatePayloadFromReq(req));
```
DFurnes: Just a thought – I've generally seen Lodash/Underscore named as `_` for convenience (or specifically importing needed methods, a la `const { extend } = require('underscore');`. Saves a little typing!

aaronschachter: Samesies. This is consistent with how we use define it in all gambit apps, but I wouldn't making this change across the board some day.

PR #412, opened 2018-10-03
Adds voting plan macros for attendingWith and timeOfDay [pr]

aaronschachter:
#### What's this PR do?

Adds new voting plan topics and macros to save who user plans to vote with, and what time of day.

* Completes DRY TODO by referencing topic helper config for hardcoded Rivescript topic changes

* Adds TODO comment about sending voting plan fields as a text post

* Upon staging deploy, begin a voting plan conversation by sending `make a plan` and answering with a `voting` response. 

* Verify all voting plan fields are saved to your user (will update https://github.com/DoSomething/gambit-admin/pull/73 accordingly)

* Sanity check that support and subscription status updating macros work as expected per DRY cleanup

I've been wondering about test coverage for all of these hardcoded config changes; it sounds like the ideal would be writing integration tests that walk through a voting plan conversation via `POST /v2/messages?origin=test` requests and [inspecting the template of the outbound message in the response.](https://github.com/DoSomething/gambit-conversations/blob/master/documentation/endpoints/messages.md#custom).. although ideally we'd need to check user status as well (which we could inspect if we included a `user` in the `response.data` of the `POST /messages` request.

#### Relevant tickets

Completes Northstar integration task in https://www.pivotaltracker.com/story/show/160824069

#### Checklist
- [ ] Tested on staging.


brain/topics/askVotingPlanStatus.rive
```diff
@@ -4,6 +4,8 @@
 // Current options are A) Yes B) No C) Already voted D) Can't vote
 // @see config/lib/macro
```
rapala61: Minor path update: `config/lib/helpers/macro`

lib/middleware/messages/member/macro-reply.js
```diff
@@ -11,6 +11,13 @@ module.exports = function replyMacro() {
       if (!macro.text) {
         return next();
       }
+
+      // TODO: If this is a votingPlanMethodOfTransport macro, we want to create a text post
```
rapala61: Nitpick: Multiple line comments usually use the /** */ form.

rapala61:
Re: testing.

Integration tests sound like a good approach to test the hard coded responses. The verbose way would be to go through all the flow in the same test, this way we could assert that each step responds with the correct replies. Alternatively, we could seed a conversation in the proper step right before each step and assert the response.

We can use `nock` to mock user's in different statuses and voting plan values.

PR #419, opened 2018-10-16
Rogue voting plans part 2: one per user [pr]

aaronschachter:
#### What's this PR do?

Checks whether a voting plan post exists in Rogue before creating, to prevent creating multiple Rogue voting posts for a user.

Verify that another round of "yes" response to a `askVotingStatusPlan` broadcast (e.g. `4LJn0P1LeoUauC0CkESYKE`) does not result in an additional Rogue post (can 👀 your user's Gambit Admin signups tab to verify)

A future iteration could execute a `PATCH /posts/:id` request to update the voting plan post with the user's latest voting plan responses -- but we don't have any current plans to send users multiple `askVotingPlanStatus` broadcasts in production. This PR goes for minimal changes, as implementing this would require updating [gateway-js](https://github.com/dosomething/gateway-js) to additionally support patch requests for Rogue posts.

#### Relevant tickets

Completes final dev task in https://www.pivotaltracker.com/story/show/160824069

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -17,9 +17,9 @@ const votingPlanPostConfig = config.posts.votingPlan;
  */
 function createVotingPlan(user, source) {
   const votingPlan = {
-    attending_with: user.voting_plan_attending_with,
-    method_of_transport: user.voting_plan_method_of_transport,
-    time_of_day: user.voting_plan_time_of_day,
+    attending_with: user[config.fields.votingPlanAttendingWith.name],
+    method_of_transport: user[config.fields.votingPlanMethodOfTransport.name],
+    time_of_day: user[config.fields.votingPlanTimeOfDay.name],
   };
   const payload = {
     northstar_id: user.id,
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -40,7 +40,7 @@ function createVotingPlan(user, source) {
  */
 async function createVotingPlanIfDoesntExist(user, source) {
   const userId = user.id;
-  const votingPlan = await module.exports.getVotingPlan(user);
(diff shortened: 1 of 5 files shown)


lib/helpers/user.js
```diff
@@ -33,6 +28,57 @@ function createVotingPlan(user, source) {
...
+}
+
+/**
+ * @param {Object} user
+ * @return {Promise}
+ */
+function fetchVotingPlan(user) {
+  const query = module.exports.getFetchVotingPlanQuery(user);
+  logger.debug('rogue.getPosts', { query });
+  return rogue.getPosts(query).then(res => res.data[0]);
```
rapala61: What do you think of using try/catch and `await` here and returning `res.data[0]`?

lib/helpers/user.js
```diff
@@ -33,6 +28,57 @@ function createVotingPlan(user, source) {
...
+ * @param {Object} user
+ * @return {String}
+ */
+function getVotingPlanDescription(user) {
+  const votingPlan = {
+    attending_with: user[config.fields.votingPlanAttendingWith.name],
+    method_of_transport: user[config.fields.votingPlanMethodOfTransport.name],
+    time_of_day: user[config.fields.votingPlanTimeOfDay.name],
+  };
+  return JSON.stringify(votingPlan);
```
rapala61: Wondering if it would be better to stringify this [here](https://github.com/DoSomething/gambit-conversations/pull/419/files#diff-0422be7d7b8af37e596dd775e1f5d172R22) rather than in the return. This way it could potentially be used in other cases in the future without having to parse it.

PR #420, opened 2018-10-16
Refactors as fetchOrCreateVotingPlan

aaronschachter:
#### What's this PR do?

Refactors `createVotingPlanIfDoesntExist` as `fetchOrCreateVotingPlan`, and logs the found-or-created voting plan for debugging.

Verify expected behavior for creating a new voting plan post (by deleting a previous voting plan if one exists already), and verify a second post is not created upon completing another voting plan flow.

Also sends JSON to the [`details` property](https://github.com/DoSomething/rogue/blob/master/docs/endpoints/posts.md#create-a-post), although from what I can tell in the API responses, it's not returned regardless of a value set (cc @katiecrane)

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/brain/topics/askVotingPlanStatus.rive
+++ b/brain/topics/askVotingPlanStatus.rive
@@ -1,4 +1,4 @@
-+ make a plan
++ voting plan
 - askVotingPlanStatus
 
 // Current options are A) Yes B) No C) Already voted D) Can't vote
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -16,12 +16,14 @@ const votingPlanPostConfig = config.posts.votingPlan;
  * @return {Promise}
  */
 function createVotingPlan(user, source) {
+  const details = JSON.stringify(module.exports.getVotingPlanValues(user));
   const payload = {
-    northstar_id: user.id,
-    type: votingPlanPostConfig.type,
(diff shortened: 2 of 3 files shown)


katiecrane:
@aaronschachter looks like [`details` are returned in API responses](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/PostTransformer.php#L59), but they are [gated](https://github.com/DoSomething/rogue/blob/master/app/Http/Transformers/PostTransformer.php#L54). Is that the behavior you're seeing?

aaronschachter:
@katiecrane -- I was expecting the `details` to appear in the post since I was including it in executing a `POST /posts` request. Per further investigaiton in #435, seems that passing `details` in a `POST /posts` request  sets `details` on the signup, not the post -- which seems to contradict advertised behavior in the [docs](https://github.com/DoSomething/rogue/blob/master/docs/endpoints/posts.md#create-a-post).

katiecrane:
@aaronschachter hm that's definitely not expected. Including `details` when creating a post should be setting it there. We also have [a test for this](https://github.com/DoSomething/rogue/blob/d7ddd36c8489dc80ce167d24319e9983afd046da/tests/Http/PostTest.php#L55-L102) so I'm confused about what could be going on here. cc: @DoSomething/team-bleed

PR #421, opened 2018-10-16
Adds additional votingPlan topic triggers

aaronschachter:
#### What's this PR do?

Branching from #420, this PR adds additional Rivescript triggers and arrays to parse the option values within the multiple choice questions in the Voting Plan topics as valid responses (e.g. If user responds with "my aunt", execute the `votingPlanStatusAttendingWithFamily` macro.

Also cleans up voting plan trigger definitions, grouping the multiple choice `a`, `b`, `c` triggers separately and removing redirects, to make it easier to switch around the a/b/c values if necessary.

I've been spot checking by manually testing an `askVotingPlanStatus` and answering with values that appear within the question (or with substitutions I've been adding).  Ideally we should be adding integration tests for `POST /messages?origin=twilio` that verify the message macros set as expected in the `response.data.messages`.

I've been wondering whether Rivescript trigger is less cumbersome than a regex for some of these arrays like `family` and `friends` that have a lot of values between the singular and plurals, but think it's safe to keep this as is now as we're looking to QA with staff early next week.

#### Relevant tickets

Fixes https://www.pivotaltracker.com/story/show/161248123

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/brain/index.rive
+++ b/brain/index.rive
@@ -66,10 +66,11 @@
  *
  * Note: although Rivescript doesn't throw an error, using an array with underscore in its name in a 
  * trigger will fail to match... which is why some of these have names like 'publictransport'.
- * Rivescript does  throw an error if an uppercase letter appears within a trigger.
+ * Rivescript does throw an error if an uppercase letter appears within a trigger.
  * @see https://github.com/aichaos/rivescript-js/issues/118#issuecomment-220413307
  */
 ! array afternoon = afternoon lunch lunchtime noon
+! array alone     = alone|me|myself|self|solo
 ! array bike      = bicycle|bike|bikin|biking|cycling
 ! array bot       = bot robot machine computer
 ! array botname   = freddie freddi alicia alysa alysha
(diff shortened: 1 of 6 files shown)


brain/index.rive
```diff
@@ -61,14 +52,39 @@
...
+! array coworkers = co-workers coworkers work
 ! array die       = die|kill myself|killing myself
   ^ commit suicide|committing suicide|commiting suicide
 ! array dosomething = dosomething|dosomethingorg|do something
 ! array drive     = car|driv|drive|drivin|driving|truck|whip
+! array evening   = even eveni evenin evening night nighttime nite
+! array family    = abeula|abeulas|abuelo|abeulos|aunt|aunts|auntie|aunties
+  ^ bro|bros|brother|brothers
+  ^ cousin|cousins|cuz|cuzzies
+  ^ dad|daddy|father|fam|family|fams|famz
```
rapala61: Can we add `familia` ?

brain/topics/askVotingPlanStatus.rive
```diff
@@ -4,28 +4,44 @@
...
-@ b
-
 + c
 - votingPlanStatusVoted
 
 + d
 - votingPlanStatusCantVote
 
++ @no [*]
+- votingPlanStatusNotVoting
```
rapala61: Should we also add triggers for potential members that express they can't vote because of their immigration status? Examples:

- `+ [*] not a citizen [*]`
- `+ [*] undocumented [*]`

PR #423, opened 2018-10-22
Inspect campaign webSignup template when sending signup messages [pr]

aaronschachter:
#### What's this PR do?

Modifies a `POST /v2/messages?origin=signup` request to set outbound message text and topic to the values found on the `config.templates.webSignup` found in a `GET /campaigns/:campaignId` Content API request.

Execute a `POST /v2/messages?origin=signup` request for your user id and campaign 2900, which is currently configured with a `webSignup` text and topic.

* Needs integration tests, potentially in future PR

* The `campaign.config.templates` property was introduced in https://github.com/DoSomething/gambit-campaigns/pull/1076 -- this work was put on hold until completion of all voting plan work.

#### Relevant tickets

#### Checklist
- [ ] Tested on staging.


lib/middleware/messages/signup/campaign-get.js
```diff
@@ -0,0 +1,36 @@
...
+
+      if (!campaign.config.id) {
+        helpers.addBlinkSuppressHeaders(res);
+        return helpers.response.sendNoContent(res, 'Campaign does not have a config.');
+      }
+
+      const webSignup = campaign.config.templates.webSignup;
+      if (!webSignup.text) {
+        helpers.addBlinkSuppressHeaders(res);
+        return helpers.response.sendNoContent(res, 'Campaign does not have a webSignup.');
```
rapala61: I would love to know more about why we are using `204`s as opposed to `422`s. `204` conveys the meaning that the POST request was successful in sending the signup SMS message, which in that case a `200` would be more appropriate IMO.

aaronschachter: The main reason is to avoid constantly logging errors for signups of campaigns that aren't configured with a web signup. If Blink was selectively executing `POST /messages?origin=signup` requests only for campaigns that have a chatbot configuration, it'd definitely make sense to send a 422 instead of a 204.

rapala61: Ah got it. If NewRelic is the main reason, I think we could try telling the NewRelic client to ignore this transaction using some rules. I think moving the check to the Blink worker, makes sense too. Either way, I think we should create a card to follow up. In the mean time, what do you think of adding some comments clarifying that `204`s are there to stop NewRelic reporting errors?

aaronschachter: Yup New Relic is the main reason, and sounds good about a follow up card -- I had wanted to move the check to the Blink worker (as Quicksilver made a similar check in our pre-TGM days), but Sergii had wanted to remove the logic from Blink. I think some reasons were that we didn't have much caching on the Gambit Content side at the time -- Quicksilver used to have be restarted to update the list of current campaigns.

I'll add some clarifying comments for the meantime -- I also need to backfill all current campaigns with `webSignup` templates in order to get this to production.

PR #425, opened 2018-10-29
WIP: Adds transition support for defaultTopicTriggers

aaronschachter:
#### What's this PR do?

Modifies `helpers.rivescript.parseRivescript` to support `defaultTopicTrigger` entries that have a transition type saved to the `response` reference field, per https://github.com/DoSomething/gambit-campaigns/pull/1094

* Also adds a `newline` Rivescript command, to avoid requiring editors to manually add a `^` character in Contentful, per https://www.rivescript.com/docs/tutorial#line-breaking

#### How should this be reviewed?

Upon a Rivescript cache refresh, the `puppetsloth` trigger that references an `autoReplyTransition` will send the entry text message, and change the conversation topic to its topic reference.

#### Any background context you want to provide?

Upon deploying this branch to production, I'll be manually backfilling any published `defaultTopicTrigger` entries with new transition types. Once all have been backfilled, we can update `helpers.rivescript.parseRivescript` to remove checking by topic type (and remove the `autoReply`, `textPostConfig`, and `photoPostConfig` types from the list of available entries that the `defaultTopicTrigger` type `response` reference field can contain)

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/157369418

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tests added/updated for new features/bug fixes.
- [ ] Tested on staging.

PR #433, opened 2018-11-05
Default topic trigger transitions - part 1 [pr]

aaronschachter:
#### What's this PR do?

Cherry picks some of #425 into a part 1 PR for supporting the newer transition content types in https://github.com/DoSomething/gambit-campaigns/pull/1094 as a `response` references on a `defaultTopicTrigger` entries.

* Refactors parsing a `defaultTopicTrigger` as Rivescript, so editors no longer need to include a `^` character as hack for[ Rivescript continuation commands.](https://www.rivescript.com/docs/tutorial#line-breaking). We'll need to update this message copy once we go live, but that can't happen until Part 2 coming soon. Part 2 fixes the topic changes that part 1 breaks -- as we no longer need to fetch the new topic Id in order to determine the message to send if a topic change happens via keyword.

This branch gets things almost-working. It creates functional Rivescript triggers on the default topic by either using the transition text field, or the topic's relevant start template field (which will be deprecated once we backfill all triggers with transition entries). We aim to manually swap them out 1 by 1 when this is ready to go live, as part 2 will need to support topic changes for both `response` types (transition entires and topic entries) during the migration.

This shouldn't be merged into master without a working part 2, but I'm opening as is to help keep this refactoring reviewable. Part 2's PR will be opened up against this branch .

We aim to make this change so the help text per start message can be edited per the topic change. Examples:

* For a `photoPostTransition` entry, the `text` field description reads as: "Should prompt user to send START to begin a photo post"

* For a `textPostTransition` entry, the `text` field description reads as: "Should ask user to send text to create a post from"

screen shot 2018-11-05 at 1 07 31 pm

#### Relevant tickets

#### Checklist

### DIFF
--- a/lib/helpers/rivescript.js
+++ b/lib/helpers/rivescript.js
@@ -11,6 +11,14 @@ const cacheKey = config.cacheKey;
 
 /**
  * @param {String} text
+ * @return {String}
+ */
+function appendEscapedLinebreak(text) {
+  return `${text}\\n`;
+}
+
+/**
+ * @param {String} text
  * @param {String} topicId
  * @return {String}
  */
@@ -158,7 +166,7 @@ function parseReplyRivescriptLines(replyText, newTopicId) {
 
     const isLastLine = index === replyTextByLine.length - 1;
     // If this isn't the last line, escape the line separator to add a double newline.
-    const content = isLastLine ? text : `${text}\\n`;
+    const content = isLastLine ? text : module.exports.appendEscapedLinebreak(text);
     // If trigger contains a topic, append a topic tag to the end of our return Rivescript.
     // @see https://www.rivescript.com/docs/tutorial#topics
     const needsTopicTag = newTopicId && isLastLine;
@@ -261,6 +269,7 @@ function parseAskYesNoResponse(messageText) {
 }
 
 module.exports = {
+  appendEscapedLinebreak,
   appendTopicTag,
   fetchRivescripts,
   formatRedirectRivescript,
--- a/lib/helpers/rivescript.js
+++ b/lib/helpers/rivescript.js
@@ -155,15 +155,12 @@ function parseReplyRivescriptLines(replyText, newTopicId) {
    *
    * NOTE: We're assuming each line break should be a double line, as we consistently use this to
    * add a "paragraph" for the longer messages we send (eg. a startPhotoPost).
+   * Because we're splitting by the newline character, we want to remove any empty strings from the
+   * result, we'll account for this by appending an escaped line break to the previous non-empty
+   * line (as long as it's not the last line, where we don't need a trailing linebreak).
    */
-  const replyTextByLine = replyText.split(config.separators.line);
-  const lines = replyTextByLine.map((text, index) => {
-    // If null, it's from a double line separator, which we accounted for in the previous item by
-    // appending an escaped line separator.
-    if (!text) {
-      return null;
-    }
-
+  const replyTextByLine = underscore.without(replyText.split(config.separators.line), '');
+  return replyTextByLine.map((text, index) => {
     const isLastLine = index === replyTextByLine.length - 1;
     // If this isn't the last line, escape the line separator to add a double newline.
     const content = isLastLine ? text : module.exports.appendEscapedLinebreak(text);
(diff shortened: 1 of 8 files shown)


lib/helpers/rivescript.js
```diff
@@ -77,55 +82,74 @@ function formatRivescriptLine(operator, value) {
...
+   * We can't stream strings that contain newline characters, as the parser complains that we're
+   * sending invalid Rivescript. Handle line breaks by split by newline and parsing as reply and
+   * newline commands.
+   * @see https://www.rivescript.com/docs/tutorial#line-breaking
+   *
+   * NOTE: We're assuming each line break should be a double line, as we consistently use this to
+   * add a "paragraph" for the longer messages we send (eg. a startPhotoPost).
+   */
+  const replyTextByLine = replyText.split(config.separators.line);
+  return replyTextByLine.map((text, index) => {
```
rapala61: I think there is enough logic from this line down to move to it's own function. Also, adding a few tests for `formatReplyRivescript` (if you move to own function, then that function) could help visualize what the function's outcome is given a few input cases.

aaronschachter: Good call! I had meant to do this, but wanted to switch gears into getting these functional in #434. I'm happy to add in this branch.

lib/helpers/rivescript.js
```diff
@@ -50,82 +67,118 @@ function fetchRivescripts() {
...
 
 /**
- * Modifies the reply property if given defaultTopicTrigger is a changeTopic macro.
+ * Returns Rivescript commands from defaultTopicTrigger.
  *
  * @param {Object} defaultTopicTrigger
- * @return {Object}
+ * @return {String}
  */
 function parseRivescript(defaultTopicTrigger) {
```
rapala61: I've spent some time in this function. Since I'm not fluent in Rivescript. I would have loved for it to have more comments as to why it does certain things, for example:

```
function parseRivescript(defaultTopicTrigger) {
  if (!defaultTopicTrigger) {
    throw new Error('parseRivescript cannot parse falsy defaultTopicTrigger');
  }

  // Start building rivescript trigger
  // + trigger\n
  const lines = [module.exports.formatTriggerRivescript(defaultTopicTrigger.trigger)];
  // @ redirect\n
  if (defaultTopicTrigger.redirect) {
    lines.push(module.exports.formatRedirectRivescript(defaultTopicTrigger.redirect));
    // + trigger\n@ redirect\n
    return module.exports.joinRivescriptLines(lines);
  }

  const topic = defaultTopicTrigger.topic;
  const topicId = topic && topic.id ? topic.id : null;
  let replyText = defaultTopicTrigger.reply;

  // Triggers that directly reference a topic entry won't have a reply set.
  // TODO: Remove this logic once they are updated to reference transition entries instead.
  if (!replyText && topicId) {
    replyText = helpers.topic.getStartTemplateText(topic);
  }

  // Array w/ reply rivescript tokens
  const replyLines = module.exports.formatReplyRivescript(replyText, topicId);

  // - reply text\\n^ continued reply text\\n^ continued las line with topic{topic=atopicid}
  return module.exports.joinRivescriptLines(lines.concat(replyLines));
}
```

lib/helpers/rivescript.js
```diff
@@ -50,82 +67,118 @@ function fetchRivescripts() {
...
+ *
+ * @param {String} text
+ * @param {String} newTopicId
  * @return {Array}
  */
-function getReplyRivescript(triggerText, replyText) {
-  const replyLine = module.exports.formatRivescriptLine(config.commands.reply, replyText);
-  return module.exports.getRivescriptFromTriggerTextAndRivescriptLine(triggerText, replyLine);
-}
+function parseReplyRivescriptLines(replyText, newTopicId) {
```
rapala61: This function is very nicely commented 💯

lib/helpers/rivescript.js
```diff
@@ -50,82 +67,118 @@ function fetchRivescripts() {
...
-  }
+    const isLastLine = index === replyTextByLine.length - 1;
+    // If this isn't the last line, escape the line separator to add a double newline.
+    const content = isLastLine ? text : module.exports.appendEscapedLinebreak(text);
+    // If trigger contains a topic, append a topic tag to the end of our return Rivescript.
+    // @see https://www.rivescript.com/docs/tutorial#topics
+    const needsTopicTag = newTopicId && isLastLine;
+
+    return {
+      operator: index === 0 ? config.commands.reply : config.commands.continuation,
```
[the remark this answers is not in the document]
aaronschachter: 🦅 👀 ! 👍

PR #434, opened 2018-11-05
Default topic trigger transitions - part 2 [pr]

aaronschachter:
#### What's this PR do?

Picks up where #433 left off to restore changing topic by default topic trigger. Because we're checking [for a topic tag within the Rivescript reply](https://www.rivescript.com/docs/tutorial#topics) to determine whether the conversation topic should change, we can deprecate the changeTopic macro and middleware by refactoring the `executeTopicChangeMacro` as `executeRivescriptTopicChange`.

Verify all default topic triggers with topic campaigns change topic and post a signup accordingly. Can also verify that our dev trigger, `puppetsloth`, which references a transition correctly sends the reply and changes topic correctly.

When this is deployed to prod, we can manually "import"  each trigger response referencing a topic entry by creating a corresponding transition entry and updating the reference.

#### Relevant tickets

Upon deploy and manual migration, this will close out https://www.pivotaltracker.com/story/show/157369418

#### Checklist
- [x] Tested on staging.


aaronschachter:
Note: as is, this branch requires a Rivescript cache clear upon deploy, as older versions of the cache won't include the topic tags added in #433. Ideally once #433 and this PR is approved, this PR should be edited to merge to `master` before merge.

lib/helpers/request.js
```diff
@@ -142,13 +148,31 @@ function getCampaignActivityPayload(req, broadcastId = null) {
...
+  const defaultTopicId = helpers.topic.getDefaultTopicId();
+  const rivescriptTopicId = isRivescriptTopic ? req.currentTopicId : defaultTopicId;
+
+  const reply = await helpers.rivescript
+    .getBotReply(req.userId, rivescriptTopicId, req.inboundMessageText);
+
+  // Check whether this Rivescript reply is changing current topic.
+  let newTopicId = null;
+  // If we're in a Rivescript topic, change when current and reply topics don't match.
+  if (isRivescriptTopic && reply.topicId !== req.currentTopicId) {
```
rapala61: From line 164 to 169, feels like it could be easier to read if we use an `||` and semantic variables:

```
  const isDifferentRivescriptTopic = isRivescriptTopic && (reply.topicId !== req.currentTopicId);
  const isNotDefaultTopic = reply.topicId !== defaultTopicId;

  // If we're in a Rivescript topic, change when current and reply topics don't match.
  // Otherwise change when the reply topic isn't the default we passed.
  if (isDifferentRivescriptTopic || isNotDefaultTopic) {
    newTopicId = reply.topicId;
  }
```

lib/helpers/rivescript.js
```diff
@@ -212,11 +212,7 @@ async function getBotReply(userId, topicId, messageText = '') {
    * triggers, e.g. passing "no text found" could trigger a "no" trigger in an askYesNo.
    */
   const userMessageText = helpers.util.containsAlphanumeric(messageText) ? messageText : 'catchAll';
-  // If Rivescript bot receives a message in a topic that it doesn't know about, it logs a
-  // "User x was in an empty topic Y" message to the console. This avoids that message.
-  const rivescriptTopicId = helpers.topic
-    .isRivescriptTopicId(topicId) ? topicId : helpers.topic.getDefaultTopicId();
-  return rivescript.getBotReply(userId, rivescriptTopicId, userMessageText);
+  return rivescript.getBotReply(userId, topicId, userMessageText);
```
rapala61: I see that you took out this `rivescriptTopicId` check and moved it into `helpers.request.getRivescriptReply` to prevent `"User x was in an empty topic Y"`. If I understand this logic correctly. Since the risk of sending a topic that the user is not in is always there regardless, would it make sense to default `rivescript.getBotReply` to get the `defaultTopicId` reply if the `topicId` passed was not the current topic?

rapala61: In other words, would it be better to move that `rivescriptTopicId ` check inside `rivescript.getBotRepl` instead of `helpers.request.getRivescriptReply`?

aaronschachter: I moved the check into  `helpers.request.getRivescriptReply` in attempt to clarify the logic for determining whether the topic has changed -- which happens after we call `helpers.rivescript.getBotReply`, since we're now sending the  `defaultTopicId` when in a non-Rivescript topic as the param to `helpers.rivescript.getBotReply`.

PR #436, opened 2018-11-08
Fixes bug when topic changes to random

aaronschachter:
#### What's this PR do?

Fixes bug introduced in #434 by adding `changeTopic` to the list of topic templates. The bug occurs when user's topic is set to random. If the next message from the user doesn't match any triggers, Gambit interprets the inbound message as continuing the current topic. Because the topic is random, Gambit finds whatever topic the user was in last, sets it, and posts the inbound message as campaign activity.  Example screen shot is below. 

* This PR adds a `getTransitionTemplateName` function to the topic helper to restore the original template names expected within conversation: `startPhotoPost`, `askText`, and now a new template `autoReplyTransition` -- removing the `changeTopic` template name completely. 

* Refactors `executeRivescriptTopicChange` request helper as `executeInboundTopicChange`, leaving TODO's about DRY in the `askYesNo` catchAll middleware

Can test campaign keywords, and test transition from TMI flow into random. Upon finishing the TMI flow, you should be asked to continue the last campaign you were in (or get a `noCampaign` reply if your conversation doesn't have a campaign Id set)

Example of the bug:

screen shot 2018-11-08 at 1 22 14 pm

#### Relevant tickets
#434 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/config/lib/helpers/topic.js
+++ b/config/lib/helpers/topic.js
@@ -1,14 +1,35 @@
 'use strict';
 
+// TODO: Merge template config here to DRY.
+const templateConfig = require('./template');
+
+const topicTemplates = templateConfig.templatesMap.topicTemplates;
+
 module.exports = {
   types: {
-    askVotingPlanStatus: 'askVotingPlanStatus',
-    askYesNo: 'askYesNo',
-    autoReply: 'autoReply',
-    photoPostConfig: 'photoPostConfig',
-    textPostConfig: 'textPostConfig',
+    askVotingPlanStatus: {
+      type: 'askVotingPlanStatus',
+    },
+    askYesNo: {
+      type: 'askYesNo',
+    },
+    autoReply: {
+      type: 'autoReply',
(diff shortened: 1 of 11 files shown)


config/lib/helpers/topic.js
```diff
@@ -1,14 +1,35 @@
...
+    askYesNo: {
+      type: 'askYesNo',
+    },
+    autoReply: {
+      type: 'autoReply',
+      transitionTemplate: topicTemplates.autoReplyTransition,
+    },
+    photoPostConfig: {
+      type: 'photoPostConfig',
+      transitionTemplate: topicTemplates.startPhotoPost,
```
rapala61: Minor: Could we add comments on what the correct transitions will be after backfilling?

aaronschachter: We'll still want these set to determine what the template name should be for transition messages (e.g. if a transition is linking to a photoPostConfig, we want to send the transition message with template set to `startPhotoPost.

PR #437, opened 2018-11-12
Cleanup askYesNo catchAll

aaronschachter:
#### What's this PR do?

This PR completes TODOs from #436 to refactor the `catchAll-askYesNo` middleware to use the new `executeInboundTopicChange` request helper function to DRY.

Verify expected behavior for responding to an `askYesNo` -- upon topic change, a signup should be created if the new topic has a campaign and signup doesn't exist for user. e.g. [5q56wA4uJ2SoAi6uAS4uYK](https://gambit-admin-staging.herokuapp.com/broadcasts/5q56wA4uJ2SoAi6uAS4uYK)

#### Relevant tickets
#436 

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/helpers/request.js
+++ b/lib/helpers/request.js
@@ -4,8 +4,6 @@ const helpers = require('../helpers');
 const gambitCampaigns = require('../gambit-campaigns');
 const logger = require('../logger');
 const analytics = require('./analytics');
-const UnprocessableEntityError = require('../../app/exceptions/UnprocessableEntityError');
-
 const config = require('../../config/lib/helpers/request');
 
 /**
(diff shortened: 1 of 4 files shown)


aaronschachter:
This branch is creating 2 signups when i say yes to the broadcast, which is odd. Investigating... the 2nd signup may be from a reply within the topic.

screen shot 2018-11-12 at 12 38 50 pm

aaronschachter:
The double signup reported in this PR occurs when: 

* 1 - User joins a topic with a campaign - a signup is posted without a run per #435
* 2 - If the new topic collects a text or photo post and sends a message that doesn't match a Rivescript trigger, it's interpeted as a campaign post and sent over to the Gambit Campaigns `/POST campaignActivity` endpoint. 

Checked with @katiecrane as I wasn't expecting to be able to create 2 signups for a user and campaign per deprecating campaign runs, but Rogue is still checking for duplicates based on the run id:
> we still are checking for dupes on user id, campaign id, AND run id: https://github.com/DoSomething/rogue/blob/master/app/Http/Controllers/SignupsController.php#L67-L68

As a result, I've altered the signup lookup in Gambit Campaigns to match the logic in this repo in https://github.com/DoSomething/gambit-campaigns/pull/1096. As I mentioned in that PR, ideally I'd like to move all signup and post creation into this repo to deprecate the gambit campaigns `/POST campaignActivity` and  Mongo database entirely (the DB is solely used for collecting multiple fields we send as a photo post to Rogue).

DFurnes:
> As I mentioned in that PR, ideally I'd like to move all signup and post creation into this repo to deprecate the gambit campaigns /POST campaignActivity and Mongo database entirely (the DB is solely used for collecting multiple fields we send as a photo post to Rogue).

Nice! That seems like a good simplification. 👌

aaronschachter:
If/when we do move over all Rogue requests to this repo, ideally we could deprecate the gambit-campaigns repo entirely by moving the broadcasts, campaigns, defaultTopicTriggers, and topics endpoints that parse and cache Contentful queries into GraphQL.

DFurnes:
Oh man, that's the dream!! 🎉

PR #450, opened 2018-11-30
Feature/more unsubscribe triggers

rapala61:
#### What's this PR do?
- Added more unsubscribed triggers:
  - `fuck you`
  - `spam`
  - `do not text`
  - `please do not send me these texts`
- A bit of cleanup

- 👀 

#### Relevant tickets
Fixes Pivotal [Card#160662089](https://www.pivotaltracker.com/story/show/160662089)

#### Checklist
- [ ] Tested on staging.


brain/index.rive
```diff
@@ -118,7 +118,8 @@
 ! array social    = social media|facebook|twitter|instagram|social
 ! array unsubscribe = unsubscribe|cancel
   ^ leave|remove|quit|end|exit|opt out|optout|discontinue|take me off
-  ^ atop|dtop|sotp|srop|stip|stop|stop all|stopp|stopall|stp
+  ^ atop|dtop|sotp|srop|stip|stop|stop all|stopp|stopall|stp|fuck you
```
aaronschachter: Instead of defining the f-bomb here, what do you think about adding a new [hardcoded trigger](https://github.com/DoSomething/gambit-conversations/blob/3.13.0/brain/macros.rive#L33) with a wildcard to catch anything that starts with one?

```
+ fuck [*]
- @stop
```

brain/index.rive
```diff
@@ -118,7 +118,8 @@
...
+  ^ spam|do not text|please do not send me these texts
```
aaronschachter: - `spam`--  definitely seems like a great addition for  `@unsubscribe` .

- `do not text` -- Instead of adding this here, can we add `send` as an additional substitution [in `@messaging`](https://github.com/DoSomething/gambit-conversations/blob/3.13.0/brain/index.rive#L107) to catch this via [`+ @please do not @messaging [*]`](https://github.com/DoSomething/gambit-conversations/blob/3.13.0/brain/macros.rive#L45)? 

- `please do not send me these texts` -- I don't think it works when you try to make a substitutions optionals (e.g. `+ [@please] do not @messaging`), so I'd think adding a `+ do not @messaging [*]` hardcoded trigger would catch `do not text` and all of the variations:

```
+ @please do not @messaging [*]
@ stop

+ do not @messaging [*]
@ stop
```

brain/index.rive
rapala61: Well that makes a lot of sense 😆👍.

PR #459, opened 2019-01-15
Adds more error codes to consider member undeliverable 

rapala61:
#### What's this PR do?
Adds new error codes to the undeliverableErrors config. This way if Twilio reports messages failing due to any of these errors, the member's `sms_status` will be set to `undeliverable`.

- 👀 
- All tests should pass

#### Relevant tickets
Fixes Pivotal [Card#161688910](https://www.pivotaltracker.com/story/show/161688910)

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/middleware/messages/update/params.js
+++ b/lib/middleware/messages/update/params.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const helpers = require('../../../helpers');
+const logger = require('../../../logger');
 
 module.exports.hasDeliveryMetadata = function hasDeliveryMetadata(body) {
   return body.metadata && body.metadata.delivery;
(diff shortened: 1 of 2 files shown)


lib/middleware/messages/update/params.js
```diff
@@ -18,6 +19,11 @@ module.exports.parseDeliveryMetadataProperties = function parseDeliveryMetadataP
       failureData,
       undeliverableError: req.undeliverableError,
     });
+    // Expose to monitor message delivery failures
+    logger.info('Outbound message delivery failure', {
+      code: req.deliveryFailureData.code,
+      messageId: req.messageId,
+    });
```
aaronschachter: Great call! Noticed the Papertrail Slack add-on 💯

PR #464, opened 2019-01-23
Fixes the re-sending of retried outbound messages to Twilio [pr]

### LINKED ISSUE #331: Double messaging members
_BUG_
-------------------
#### Current Behavior
- Under some circumstances. Members get 2 messages instead of 1. Twilio is reporting 2 separate "delivered" events for the same member. The second delivered message is not stored in G-Conversations.

#### Desired Behavior
- Members should only get 1 message under any circumstances. 

#### Steps to Replicate
- WIP

#### Why This Matters
- It's breaking Blink.
- It's sending members 2 messages so it's annoying and double the expected cost.

- This bug was reported in Blink first. Most of the sleuthing results are stored there: https://github.com/DoSomething/blink/issues/203

rapala61:
#### What's this PR do?
This PR inspects loaded outbound messages on retry. When a loaded outbound message's `platformMessageId` exists means it was successfully sent to Twilio.

When G-Conversations receives a retry request and the loaded outbound message was sent successfully to Twilio. It sets the `suppressOutbound` property in the `req` object to `true`. This way it won't be sent to Twilio again.

- 👀 
- All tests should pass

Under a heavy broadcast load we may experience an increase in [H12 - timeout](https://devcenter.heroku.com/articles/error-codes#h12-request-timeout) errors. Some of these requests actually finish being processed internally by the Node.js process but go over the hard 30 seconds limit imposed by the Heroku routers. These errors are reported back to Blink, which interprets that that the requests were not successfully processed and retries them.

#### Relevant tickets
Fixes #331 
Fixes Pivotal [Card#161604351](https://www.pivotaltracker.com/story/show/161604351)

#### Checklist
- [x] Tested on staging.


rapala61:
Blocked by: https://github.com/DoSomething/gambit-conversations/pull/465 since I need to merge the GraphQL changes introduced there to successfully add the integration tests for `/api/v2/messages?origin=twilio` route.

rapala61:
That was the fastest unblock ever. Thanks @aaronschachter !

PR #467, opened 2019-01-25
Fetch web signup confirmation templates from GraphQL [pr]

aaronschachter:
#### What's this PR do?

Sources signup message content from the GraphQL `webSignupConfirmations` query added in https://github.com/DoSomething/graphql/pull/44, instead of the Gambit Content `GET /campaigns/:id` endpoint.  Per #466, this PR is the last trace of the Gambit Content API in this app.

Verify signups requests are sent for campaign IDs 8226 and 5646.

Once https://github.com/DoSomething/graphql/pull/47 is deployed to production, we can add a check to prevent sending confirmations for closed campaigns (this may happen when importing signups for campaigns that have ended)

#### Checklist
- [ ] Tested on staging.


lib/helpers/campaign.js
```diff
@@ -1,14 +1,38 @@
...
-const gambitContent = require('../gambit-content');
+const graphql = require('../graphql');
+const logger = require('../logger');
+const cache = require('./cache').webSignupConfirmations;
+
+/**
+ * @return {Promise}
+ */
+async function fetchWebSignupConfirmations() {
+  return cache.set(await graphql.fetchWebSignupConfirmations());
```
rapala61: What do you think of creating a test for the scenario where `graphql.fetchWebSignupConfirmations()` fails ? I'm not sure what the caching behavior would be if GraphQL is unavailable.

lib/helpers/campaign.js
```diff
@@ -1,14 +1,38 @@
...
+}
+
+/**
```
rapala61: Just for consistency, I think we could add the descriptions to these function comments?

lib/middleware/messages/signup/web-signup-confirmation-get.js
```diff
@@ -21,19 +16,19 @@ module.exports = function getCampaign() {
...
+      //  This sanity check is useful if importing signups from campaigns that have ended.
+      if (helpers.campaign.isClosedCampaign(template.campaign)) {
+        return helpers.sendErrorResponse(res, new UnprocessableEntityError('Campaign has ended.'));
       }
 
-      helpers.request.setOutboundMessageText(req, webSignup.text);
-      helpers.request.setOutboundMessageTemplate(req, webSignup.template);
-      helpers.request.setTopic(req, webSignup.topic);
+      helpers.request.setOutboundMessageText(req, template.text);
+      helpers.request.setOutboundMessageTemplate(req, 'webSignupConfirmation');
```
rapala61: We could move the string `webSignupConfirmation` to a config so we can also use it in tests without hardcoding them.

app/routes/messages/signup.js
```diff
@@ -14,16 +14,16 @@ const getUserMiddleware = require('../../../lib/middleware/messages/user-get');
 const validateOutboundMessageMiddleware = require('../../../lib/middleware/messages/message-outbound-validate');
 const getConversationMiddleware = require('../../../lib/middleware/messages/conversation-get');
 const createConversationMiddleware = require('../../../lib/middleware/messages/conversation-create');
-const getCampaignMiddleware = require('../../../lib/middleware/messages/signup/campaign-get');
+const getWebSignupConfirmationMiddleware = require('../../../lib/middleware/messages/signup/web-signup-confirmation-get');
```
rapala61: This new naming makes so much more sense! 💯

test/unit/lib/middleware/messages/message-outbound-create.test.js
```diff
@@ -64,16 +64,45 @@ test('createOutboundMessage calls next if req.outboundMessage exists', async (t)
...
+  t.context.req.outboundMessageTemplate = stubs.getRandomWord();
   const middleware = createOutboundMessage(configStub);
 
   // test
   await middleware(t.context.req, t.context.res, next);
-  t.context.req.conversation.createAndSetLastOutboundMessage.should.have.been.called;
+  t.context.req.conversation.createAndSetLastOutboundMessage.should.have.been.calledWith(
+    configStub.messageDirection,
+    t.context.req.outboundMessageText,
+    t.context.req.outboundMessageTemplate,
```
rapala61: A nit. I would suggest you save the value of `stubs.getRandomWord()` and then compare here. The way the test is comparing against `t.context.req.outboundMessageTemplate` could potentially hide a mutation bug.

aaronschachter:
I've deployed https://github.com/DoSomething/gambit-slack/pull/45 that allows testing signup messages via a new `signup` Slack command:

screen shot 2019-01-28 at 4 52 37 pm

PR #468, opened 2019-01-29
Refactors first iteration of the NewRelic `noticeError` implementation [pr]

rapala61:
#### What's this PR do?
It adds a decorated version of each method: `helpers.sendErrorResponse` and `helpers.sendErrorResponseWithNoRetry`. These versions will call `newrelic.noticeError` and log the error metadata in them.

*Usage*: We can use the `helpers.errorNoticeable.sendErrorResponseWithNoRetry` method to send a no reply response with an error, while logging the error metadata in NewRelic. Using the `helpers.sendErrorResponseWithNoRetry` method works the same but won't log the metadata to NewRelic.

- 👀 
- All tests should pass

We were sending all error metadata to NewRelic, but has become very noisy.

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/bertly/index.js
+++ b/lib/bertly/index.js
@@ -6,6 +6,7 @@ const linkify = require('linkify-it')({
   fuzzyEmail: false,
 });
 
+const analyticsHelper = require('../helpers/analytics');
 const logger = require('../logger');
 const restClient = require('./rest-client');
 const config = require('../../config/lib/bertly');
@@ -92,6 +93,8 @@ async function parseLinksInTextIntoRedirects(text = '') {
 
     return parsedMessage;
   } catch (error) {
+    // Expose error metadata in NewRelic
+    analyticsHelper.addHandledError(error);
     logger.error('parseLinksInTextIntoRedirects() Error creating redirects.', {
       error: error.message,
       status: error.status,
--- a/lib/helpers.js
+++ b/lib/helpers.js
@@ -19,9 +19,9 @@ Object.keys(helpers).forEach((helperName) => {
  *
  * @param  {Object} res
  * @param  {Error} err
- * @param  {Boolean} sendSuppressHeader
+ * @param  {Boolean} retry by default, errors should be retried by Blink
  */
-module.exports.sendErrorResponse = function (res, err, sendSuppressHeader = false) {
+module.exports.sendErrorResponse = function (res, err, retry = true) {
   // Capture error in transaction
   helpers.analytics.addHandledError(err);
   const error = helpers.util.parseStatusAndMessageFromError(err);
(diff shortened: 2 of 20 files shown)


aaronschachter:
Potentially out of scope for this PR, but one thought that comes to mind while we're in these functions and the middleware that calls them is that it could be a good chance to move them out of the `helpers.js` and into [`helpers/response.js`](url)

lib/helpers.js
```diff
@@ -19,30 +19,31 @@ Object.keys(helpers).forEach((helperName) => {
...
    * We have to also relay the Suppress headers.
    */
   const errHasResponseGet = error.response && error.response.get;
-  if (sendSuppressHeader) {
+  if (!retry) {
     exports.addBlinkSuppressHeaders(res);
+  /**
+   * TODO: deprecate? this check is here because we used to relay errors from G-Campaigns
+   * but since we are deprecating it, we could remove this extra check.
+   */
```
aaronschachter: Sounds right!

lib/bertly/index.js
```diff
@@ -92,6 +93,8 @@ async function parseLinksInTextIntoRedirects(text = '') {
 
     return parsedMessage;
   } catch (error) {
+    // Expose error metadata in NewRelic
+    analyticsHelper.addHandledError(error);
```
rapala61: @aaronschachter I added this so we can hopefully get some insights into the Bertly link errors we are seeing while under broadcasts. It also means it will get noisy once they start happening. Hopefully is the productive kind of noisy.

lib/helpers.js
```diff
@@ -3,7 +3,7 @@
 const { STATUS_CODES } = require('http');
 
 const logger = require('./logger');
-const config = require('../config/lib/helpers');
+const config = require('../config/lib/helpers/response');
```
aaronschachter: 👍

lib/helpers.js
```diff
@@ -17,38 +17,27 @@ Object.keys(helpers).forEach((helperName) => {
...
  *
- * @param  {Object} res
+ * @param  {Response} res
  * @param  {Error} err
- * @param  {Boolean} sendSuppressHeader
+ * @param  {Boolean} retry by default, errors should be retried by Blink
  */
-module.exports.sendErrorResponse = function (res, err, sendSuppressHeader = false) {
-  // Capture error in transaction
-  helpers.analytics.addHandledError(err);
```
aaronschachter: Refs #478

lib/middleware/messages/user-get.js
```diff
@@ -16,6 +16,7 @@ module.exports = function fetchUser(config) {
     try {
       let user;
       /**
+       * TODO: deprecate? Are we ever using the feature of making anonymous get user requests?
```
aaronschachter: Refs #482

PR #471, opened 2019-02-04
Fetch askSubscriptionStatus replies from GraphQL

aaronschachter:
#### What's this PR do?

Removes the hardcoded macro replies for an `askSubscriptionStatus` broadcast, and refactors middleware to send replies from the GraphQL fields added in https://github.com/DoSomething/graphql/pull/49 and https://github.com/DoSomething/graphql/pull/51.

Also refactors `askSubscriptionStatus` broadcasts to no longer update a user's `sms_status` as `pending` upon receiving the broadcast, per [Slack thread](https://dosomething.slack.com/archives/C03T8SDMS/p1549318310005500). 

#### How should this be reviewed?

* Verify expected behavior for the various replies defined on an `askSubscriptionStatus` broadcast (`4qLXWRSqY0ommayY2iWCwU`) per the GraphQL `topic` query response --  see https://github.com/DoSomething/graphql/pull/51#issuecomment-460359256

Example:
screen shot 2019-02-04 at 6 38 43 pm

* Verify the `less` keyword sends the hardcoded text per https://github.com/DoSomething/gambit/pull/471#discussion_r253675066

Example:
screen shot 2019-02-04 at 7 01 47 pm

#### Any background context you want to provide?

The last broadcast type left to port over to GraphQL is `askVotingPlanStatus`. Not that we're sending one of these any time soon, but it seems worth fixing it up vs scrapping the feature from the repo entirely.  

#### Relevant tickets
Fixes https://www.pivotaltracker.com/story/show/163095523

#### Checklist
- [x] Tests added/updated for new features/bug fixes.
- [x] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love.

PR #473, opened 2019-02-06
Update askVotingPlanStatus topic replies

aaronschachter:
#### What's this PR do?

This PR fixes the `askVotingPlanStatus` topic middleware, which was broken by fetching topics from GraphQL in #461. 

Verify expected replies for a voting plan broadcast, `4LJn0P1LeoUauC0CkESYKE`. Note that answering with a `'voting'` status will send the reply to send a photo, but changes topic to one with a campaign that has ended (so the next reply will be "Sorry, this campaign is no longer available")

It was a bit confusing to be referencing this code while building out #471, so it felt like it was time to finally finish porting over all broadcast types so there isn't broken code hanging out in the repo.

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/brain/topics/askVotingPlanStatus.rive
+++ b/brain/topics/askVotingPlanStatus.rive
@@ -43,7 +43,7 @@
 - votingPlanStatusCantVote
 
 + [*]
-- invalidVotingPlanStatus
+- invalidAskVotingPlanStatusResponse
 
 < topic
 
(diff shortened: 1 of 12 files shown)


rapala61:
LGTM. The only detail is that the answer I get when actually planning my vote is not `Sorry, this campaign is no longer available`, but:

```
Sorry, I'm not sure how to respond to that.

Text Q if you have a question.
```

aaronschachter:
Fixed this odd bug that only seems to be happening on staging in https://github.com/DoSomething/gambit/pull/473/commits/69afc095d4665b05028f6b5d0dc01405d027e298, by way of debugging in #474

PR #477, opened 2019-02-15
Truncate text value when posting to Rogue

aaronschachter:
#### What's this PR do?

This PR truncates any `text` values over 256 characters to prevent [text posts over 256 chars from silently failing](https://dosomething.slack.com/archives/C1V0M6RPE/p1550251161000800).

Test that sending a reply of over 256 chars to text post broadcast `2YaK1CSXVSR4tEVetRIMh2` results in a reply message, as well as a truncated text post.

There's a separate issue for why Blink isn't retrying the 422 `The given data was invalid.` error that Gambit receives from Rogue (from what I can see in Slack, it looks like Gambit indeed does throw it), but for sake of getting this to prod sooner than later on a Friday, opening as is.

screen shot 2019-02-15 at 10 13 42 am

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const lodash = require('lodash');
 const { GatewayClient } = require('@dosomething/gateway/server');
 
 let gatewayClient;
@@ -22,7 +23,8 @@ function getConfig() {
  * @return {Promise}
  */
 function createPost(data) {
-  return module.exports.getClient().Rogue.Posts.create(data);
+  return module.exports.getClient().Rogue.Posts
+    .create(Object.assign(data, { text: lodash.truncate(data.text, { length: 256 }) }));
 }
 
 /**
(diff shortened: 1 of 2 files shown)


aaronschachter:
Verified Rogue post is truncated.

screen shot 2019-02-15 at 12 11 19 pm

aaronschachter:
Blink wasn't retrying this error because by design, it won't retry a 422 (it'd just fail every time). We weren't seeing these errors in Gambit New Relic because we're ignoring them via the `NEW_RELIC_ERROR_COLLECTOR_IGNORE_ERROR_CODES` config variable on production. The reasons for this are documented (for now?) in this [Slack thread](https://dosomething.slack.com/archives/C1V0M6RPE/p1550599590005800?thread_ts=1550596664.001400&cid=C1V0M6RPE), along with ideas for visibility into 422's from Northstar and Rogue.

PR #480, opened 2019-03-01
Pass action_id to Rogue

rapala61:
#### What's this PR do?
It passes the topic's `action_id` to Rogue when submitting text and photo posts.

- 👀 
- all tests should pass

Note: Pending tests cleanup

#### Relevant tickets
Fixes Pivotal [Card#163779560](https://www.pivotaltracker.com/story/show/163779560)

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -35,8 +35,8 @@ async function createSignup(user, campaign, signupSource, signupSourceDetails) {
  *
  * @param {Object} props - The properties used to create the photo post
  * @param {String} props.userId
- * @param {Integer} props.campaignId
- * @param {Integer} props.actionId
+ * @param {Number} props.campaignId
+ * @param {Number} props.actionId
  * @param {String} props.photoPostSource
  * @param {Object} props.photoPostValues
  * @return {Promise}
@@ -77,8 +77,8 @@ async function createPhotoPost({ userId, campaignId, actionId, photoPostSource, 
  *
  * @param {Object} props - The properties used to create the text post
  * @param {String} props.userId
- * @param {Integer} props.campaignId
- * @param {Integer} props.actionId
+ * @param {Number} props.campaignId
+ * @param {Number} props.actionId
  * @param {String} props.textPostSource
  * @param {String} props.textPostText
  * @return {Promise}
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -52,9 +52,10 @@ async function createPhotoPost({ userId, actionId, photoPostSource, photoPostVal
   if (photoPostValues.whyParticipated) {
     payload.why_participated = photoPostValues.whyParticipated;
   }
+  // log payload before fetching image.
+  logger.debug('createPhotoPost payload', { payload });
   payload[gatewayClientConfig.photoPostCreation.fileProperty] = await helpers.util
     .fetchImageFileFromUrl(photoPostValues.url);
-  logger.debug('createPhotoPost payload', { payload });
 
   return gateway.createPost(payload);
 }
--- a/lib/helpers/user.js
+++ b/lib/helpers/user.js
@@ -35,15 +35,15 @@ async function createSignup(user, campaign, signupSource, signupSourceDetails) {
  *
  * @param {Object} props - The properties used to create the photo post
  * @param {String} props.userId
- * @param {Number} props.campaignId
  * @param {Number} props.actionId
  * @param {String} props.photoPostSource
  * @param {Object} props.photoPostValues
  * @return {Promise}
  */
-async function createPhotoPost({ userId, campaignId, actionId, photoPostSource, photoPostValues }) {
+async function createPhotoPost({ userId, actionId, photoPostSource, photoPostValues }) {
   const payload = {
     northstar_id: userId,
+    action_id: actionId,
     quantity: photoPostValues.quantity,
     source: photoPostSource,
     text: photoPostValues.caption,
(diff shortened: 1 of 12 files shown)


lib/helpers/user.js
```diff
@@ -51,32 +52,70 @@ async function createPhotoPost(user, campaign, photoPostSource, photoPostValues)
...
     text: textPostText,
     type: config.posts.text.type,
   };
+
+  if (actionId) {
+    /**
+     * WARNING
+     * Posts for topics in Contentful will use the "default"
+     * backfilled action in Rogue if the "Campaign Action Id" field is left empty
+     * by content editors.
```
aaronschachter: Once https://github.com/DoSomething/rogue/pull/837 has deployed, why support sending a `campaignId` vs an `actionId` at all? We can just be sure to backfill any campaign topics currently used (right now, this is a small amount) make the field required in Contentful, and then remove this logic to simplify? Same comment re: photo posts below

For the sake of old (or, all) campaign topics that won't have an `actionId` set, GraphQL could potentially query Rogue to return the relevant action in case we ever query for the topic (this would happen either for conversations in older topics, or an admin viewing topics via Gambit Admin)

lib/middleware/messages/member/topics/posts/photo/post-create.js
```diff
@@ -14,17 +14,33 @@ module.exports = function createPhotoPost() {
...
       return await helpers.replies.completedPhotoPost(req, res);
-    } catch (err) {
-      // TODO: Prompt user to send a diffferent photo if Rogue returns error for image file size.
-      return helpers.sendErrorResponse(res, err);
+    } catch (error) {
+      /**
+       * TODO: Prompt user to send a different photo if Rogue returns error for image file size.
+       *
+       * Manual misconfiguration of topics can result in errors when
+       * submitting text posts to Rogue. We must notice this errors in
```
aaronschachter: Nit: these errors (vs this errors)

lib/helpers/user.js
rapala61: >Once [DoSomething/rogue#837](https://github.com/DoSomething/rogue/pull/837) has deployed, why support sending a `campaignId` vs an `actionId` at all? We can just be sure to backfill any campaign topics currently used (right now, this is a small amount) make the field required in Contentful, and then remove this logic to simplify? Same comment re: photo posts below

That's a good idea. I'll do that in Contentful and then make the updates to the PR. Thanks!

>For the sake of old (or, all) campaign topics that won't have an `actionId` set, GraphQL could potentially query Rogue to return the relevant action in case we ever query for the topic (this would happen either for conversations in older topics, or an admin viewing topics via Gambit Admin). That -- or we can run a script to update all text and photo post Contentful entries.

I'm not sure I follow 100%.  All campaign topics in conversations should be in Contentful. If I update them with their respective action ids, they would include it in the GraphQL payload right?

aaronschachter: > If I update them with their respective action ids, they would include it in the GraphQL payload right

Yup! I was rambling about programmatically updating about `textPostConfig` and `photoPostConfig` entries, but if it's a small enough subset - manual sometimes can be easiest (I know I've done it a bunch in our Contentful space)

aaronschachter: > That's a good idea. I'll do that in Contentful and then make the updates to the PR.

Once Action ID is added and required, let's disable the requirement for Campaign ID and hide it from editing. Eventually we'll want to delete these field once the deprecated reference is removed in GraphQL.

rapala61: @aaronschachter I'm going to tackle the cleanup of Campaign id in a different card just because I haven't really researched what the side effects are of this. This PR is ready for review :).

test/helpers/factories/topic.js
```diff
@@ -50,24 +56,28 @@ function getValidAutoReply() {
  * @return {Object}
  */
 function getValidPhotoPostConfig() {
-  return getValidTopic(config.types.photoPostConfig.type, {
+  const topic = getValidTopic(config.types.photoPostConfig.type, {
```
aaronschachter: Minor, but why not just an include an `actionId` parameter in our 2nd argument for `getValidTopic` to avoid the extra assignment?

test/helpers/factories/topic.js
```diff
@@ -50,24 +56,28 @@ function getValidAutoReply() {
...
  * @return {Object}
  */
 function getValidTextPostConfig() {
-  return getValidTopic(config.types.textPostConfig.type, {
+  const topic = getValidTopic(config.types.textPostConfig.type, {
     invalidAskText: getTemplate(),
     completedTextPost: getTemplate(),
     completedTextPostAutoReply: getTemplate(),
   });
+  topic.actionId = stubs.getRandomNumericId();
```
aaronschachter: Same question as `getValidPhotoPostConfig` changes

test/helpers/factories/topic.js
```diff
@@ -8,7 +8,13 @@ function getTemplate() {
 }
 
 /**
- * These topic stubs correspond to data returned from Query.topic GraphQL requests.
+ * TODO: Let's figure out a way to utilize the queries setup in config/lib/graphql.
+ * Managing this factory completely separate from where we defined the expected
+ * topic schema is a hard job!
+ *
+ * @see https://www.apollographql.com/docs/graphql-tools/mocking.html Probably?
```
aaronschachter: Sounds awesome!

test/helpers/factories/topic.js
```diff
@@ -8,7 +8,13 @@ function getTemplate() {
...
+ *
+ * These topic stubs correspond to data returned from Query.topic GraphQL requests.s
```
aaronschachter: Nit: trailing `s` typo

test/helpers/factories/topic.js
rapala61: Ah, yes. I felt it was clear that what was included in that object was just `tamplates`. What about if I change the parameter's name to `metadata` and include the `actionId` in it?

rapala61: Changed!

PR #485, opened 2019-03-08
Cleanup ahead of Adding broadcast campaign id to inbound and outbound messages

rapala61:
#### What's this WIP PR do?
Cleanup TODOs and adds comments. 

- 👀 
- tests should pass

#### Relevant tickets
Fixes Pivotal [Card#164076750](https://www.pivotaltracker.com/story/show/164076750)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/helpers/request.js
+++ b/lib/helpers/request.js
@@ -7,13 +7,13 @@ const config = require('../../config/lib/helpers/request');
 const DraftSubmission = require('../../app/models/DraftSubmission');
 
 /**
- * Sets the request topic and saves it as the request conversation topic.
+ * Sets the request topic. It updates the conversation w/ the topic if it has changed.
  *
  * @param {Object} req
  * @param {Object} topic
  * @return {Promise}
  */
-function changeTopic(req, topic) {
+function updateTopicIfChanged(req, topic) {
   module.exports.setTopic(req, topic);
   return (req.currentTopicId === topic.id) ? Promise.resolve() : req.conversation.setTopic(topic);
 }
@@ -44,7 +44,7 @@ async function executeInboundTopicChange(req, topic, signupDetails = null) {
   if (helpers.topic.hasActiveCampaign(topic)) {
     await helpers.user.fetchOrCreateSignup(req.user, topic.campaign, req.platform, signupDetails);
   }
-  return module.exports.changeTopic(req, topic);
+  return module.exports.updateTopicIfChanged(req, topic);
 }
 
 /**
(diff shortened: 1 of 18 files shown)


lib/helpers/request.js
```diff
@@ -370,7 +375,7 @@ function setMacro(req, macro) {
 function setTopic(req, topic) {
   req.topic = topic;
   analytics.addCustomAttributes({ topicId: topic.id });
-  if (topic.campaign && !req.campaign) {
```
rapala61: I'm removing this additional conditional since `req.campaign` is only set inside `setCampaign` which is called in the next line.

lib/helpers/request.js
```diff
@@ -323,7 +323,12 @@ function saveDraftSubmissionValue(req, key, value) {
  */
 function setCampaign(req, campaign) {
   req.campaign = campaign;
-  if (campaign && campaign.id && !req.campaignId) {
```
rapala61: I'm also removing this additional `&& !req.campaignId` since `req.campaignId` is only set in `setCampaignId `, called in the next line.

test/unit/lib/lib-helpers/request.test.js
```diff
@@ -516,16 +516,6 @@ test('setCampaign should inject a campaign property to req and call setCampaignI
...
-test('setCampaign should not call setCampaignId if req.campaignId', (t) => {
-  sandbox.spy(requestHelper, 'setCampaignId');
-  const campaign = campaignFactory.getValidCampaign();
-  t.context.req.campaignId = stubs.getCampaignId();
-
-  requestHelper.setCampaign(t.context.req, campaign);
-  t.context.req.campaign.should.deep.equal(campaign);
-  requestHelper.setCampaignId.should.not.have.been.called;
-});
-
```
rapala61: I removed the test above that asserted that `setCampaign` would not be called if `req.campaignId` existed. Based on my understanding of the code, this should never happen since `helpers.request.setCampaignId` is only called from within `helpers.request.setCampaign`. The outlier here is the `?origin=signup` route, which does call `helpers.request.setCampaignId` directly, but doesn't call `helpers.request.setCampaign` at all, so this test would also not be applicable in this case.

lib/helpers/request.js
```diff
@@ -323,7 +323,12 @@ function saveDraftSubmissionValue(req, key, value) {
...
+  analytics.addCustomAttributes({ campaign });
+  /**
+   * TODO: If we remove the campaign field in contentful for some topics. We will have to
+   * tell GraphQL to fetch it from Rogue based on the action Id?
```
aaronschachter: Sounds right! We'll also want to check for action.campaign.endDate eventually for error-proofing broadcasts, or preventing campaign activity for campaigns that have ended.

lib/helpers/request.js
```diff
@@ -340,17 +345,25 @@ function setDraftSubmission(req, draftSubmission) {
...
+ *
+ * TODO: I feel like this function is doing so much more than just setting the conversation
+ * should it be split into other functions?
  */
 function setConversation(req, conversation) {
   req.conversation = conversation;
+  /**
+   * `conversation.topic` is a reference to the topic of the last outbound message to the member.
+   * We consider the member to be in this "current topic", regardless of what
+   * new topic they will be transitioned to as a product of their inbound text.
```
aaronschachter: Documentation 💯 📖 🏅

lib/helpers/request.js
rapala61: I'll add this step in the card to cleanup the campaign field in Contentful.

aaronschachter:
is this ready for merge?

PR #486, opened 2019-03-11
It will associate messages w/ broadcast's campaign id for data analysis [pr]

rapala61:
#### What's this PR do?
- It stores the campaign topic broadcast's campaign id in `outbound-api-send` messages.
- Adds schema version to messages

- 👀 
- all tests should pass

what else?

#### Relevant tickets
Fixes #
Fixes Pivotal [Card#164076750](https://www.pivotaltracker.com/story/show/164076750)

#### Checklist
- [ ] Tested on staging.


lib/helpers/request.js
```diff
@@ -8,14 +8,16 @@ const DraftSubmission = require('../../app/models/DraftSubmission');
...
+ * @async
  * @param {Object} req
  * @param {Object} topic
  * @return {Promise}
  */
-function updateTopicIfChanged(req, topic) {
-  module.exports.setTopic(req, topic);
-  return (req.currentTopicId === topic.id) ? Promise.resolve() : req.conversation.setTopic(topic);
+async function updateTopicIfChanged(req, topic) {
+  if (req.currentTopicId !== topic.id) {
```
rapala61: @aaronschachter I refactored this method to set and update the topic only when it differs from the current one. We were setting regardless of comparison but were not updating conversations if they were the same, which might result in a mismatch of conversation and message. Just pointing out in case that was being done for an specific reason that I didn't take into account.

rapala61:
This breaks multi choice broadcasts. Partially this is because I am relying on injecting `type` and `broadcastId` to the broadcast object GraphQL returns. It's definitely not the best approach because it requires the [`fetchById`](https://github.com/DoSomething/gambit/blob/feature/associate-messages-to-broadcast-campaign/lib/helpers/broadcast.js#L14) to be aware of all the discrepancies.

Creating a transformer that takes care of normalizing broadcasts would help simplify the code. I'll try to fix this issue following the pattern we are using already, but it might just be worth it to invest more time to fix as I suggest above.

rapala61:
Last commit fixes the Non-topic broadcasts.

app/models/Message.js
```diff
@@ -32,15 +34,18 @@ const messageSchema = new mongoose.Schema({
   match: String,
   macro: String,
   metadata: {
-    requestId: { type: String, index: true },
-    retryCount: Number,
+    // Campaign Associated with this message, used for conversion analysis by Team Storm.
+    campaignId: Number,
```
aaronschachter: Why not just re-purpose our existing `campaignId` property? That was the idea I had behind introduction a schema property. Otherwise I believe Storm will need to make changes to check for `metadata.campaignId`, whereas they are already ingesting the `campaignId`. If we do keep `metadata.campaignId`, it'd make sense to scrap the old `campaignId` property completely, as we don't use it in code anymore (and have confirmation that Storm doesn't have a need for it).

rapala61: Great point. I did mention that this was the path I was taking in the Pivotal card. Since I didn't hear an objection from Team Storm I assumed it was ok on their end. I was thinking on removing that `campaignId` property you mentioned and leaving the `metadata.campaignId`.

rapala61: I've added a request for their input in the Pivotal card just to triple check!

lib/helpers/request.js
```diff
@@ -8,14 +8,40 @@ const DraftSubmission = require('../../app/models/DraftSubmission');
...
+      broadcastId = nestedTopic.broadcastId;
+      campaignId = nestedTopic.campaign ? nestedTopic.campaign.id : null;
+    } else {
+      // AskSubscriptionStatus would not have a nested topic so the broadcastId
+      // is the topic.id
+      broadcastId = topic.broadcastId || topic.id;
+      campaignId = topic.campaign ? topic.campaign.id : null;
+    }
+
+    await req.conversation.setLastReceivedBroadcast(broadcastId, campaignId);
```
aaronschachter: I'm having trouble following this logic as well (along with why we need the changes in `isBroadcastable`). Isn't it the case that the only time we want to set the lastReceivedBroadcast properties is in ` POST /messages?origin=broadcast` request? This `updateTopicIfChanged` function gets called from a few different spots within a member message... ah but talking through this I suppose we want to set it to `null` for processing a new inbound, which wouldn't be a reply to the last sent broadcast...

config/app/models/message.js
```diff
@@ -0,0 +1,30 @@
...
+
+function isOutbound(direction) {
+  return direction.includes('outbound');
+}
+
+function isInbound(direction) {
+  return direction === DIRECTION.inbound;
+}
+
+function isOutboundApi(direction) {
```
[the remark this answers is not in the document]
rapala61: No. Removing it. It was leftover from some other code I tried but changed in the last version of this PR.

app/models/Conversation.js
```diff
@@ -26,12 +27,17 @@ const conversationSchema = new mongoose.Schema({
     index: true,
   },
   topic: String,
-  campaignId: Number,
   lastOutboundMessage: {
     type: mongoose.Schema.Types.ObjectId,
     ref: 'Message',
   },
```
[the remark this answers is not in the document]
rapala61: It's not clear to me what you mean from the code you highlighted. I'm going to assume you are asking me if saving `lastReceivedBroadcastCampaignId` in the conversation schema makes sense.

This was the way to go before receiving clarification from the data team that they wanted only the `outbound-api-send` message to contain the `campaignId`. Prior to that, the task was to add this property to all messages. For that, saving a reference in the conversation was going to be the best way to achieve that. When I switched gears, I decided to keep the underlying idea.

I don't think it makes the schema any harder to read since the property name is consistent with the `lastReceivedBroadcastId`. That being said. Saving it in the outbound message can be achieved by setting `req.broadcastCampaignId` before creating the outbound message. I could make that change too!

lib/helpers/request.js
```diff
@@ -8,14 +8,40 @@ const DraftSubmission = require('../../app/models/DraftSubmission');
...
+  }
+  /**
+   * We could send members different broadcasts, with different ids
+   * that share the same topic id. For this reason, we would need to
+   * update the lastReceivedBroadcastId regardless of if the topic ids
+   * remain the same.
+   */
+  if (helpers.topic.isBroadcastable(topic)) {
+    let campaignId = null;
+    let broadcastId = null;
```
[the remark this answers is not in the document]
rapala61: Yeah, this really tripped me up too. Remember the bug I found about [`lastReceivedBroadcast` not being saved?](https://dosomething.slack.com/archives/C2C8NLNAY/p1552409770058100). Well, to fix it, I had to inject the `type` inside the `topic` object in [broadcast responses from GraphQL](https://github.com/DoSomething/gambit/blob/feature/associate-messages-to-broadcast-campaign/lib/helpers/broadcast.js#L21). [Just like we do with topics](https://github.com/DoSomething/gambit/blob/feature/associate-messages-to-broadcast-campaign/lib/helpers/topic.js#L22).

The topic passed to `updateTopicIfChanged` can be any type of topic (broadcasts, campaign topics). When it's "broadcastable", it could be any of [these](https://github.com/DoSomething/gambit/blob/feature/associate-messages-to-broadcast-campaign/config/lib/helpers/broadcast.js#L4).

For `askMultipleChoice`, `askSubscriptionStatus`, and `askVotingPlanStatus` the `topic.id` *is* the broadcast id.

For `photoPostBroadcast`, `textPostBroadcast`, and `autoReplyBroadcast`. The `topic.id` is not the broadcast id, it's the topic we are switching to, for this reason I have to inject it in the nested topic object. Which is what is passed here.

For `askYesNo` broadcast. The `topic.id` is also the broadcast id, but it differs from the rest in that it also has a nested object that contains the campaign metadata: `saidYesTopic`.

This mess is likely better fixed with changes in GraphQL and/or adding transformers in Gambit that will normalize the schema we get to work with once we receive from GraphQL.

I hope I've done a good job explaining some context behind how `updateTopicIfChanged` is structured.

app/models/Conversation.js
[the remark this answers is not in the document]
aaronschachter: I was just highlighting the end of properties, but I'm referring to adding a `conversation.schema` property. As of this PR we'll no longer be saving a `campaignId`, similar to how we're no longer saving `message.campaignId` as per this new release where messages no longer have a `campaignId` set.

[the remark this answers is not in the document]
rapala61: Oh! I reread with this context. I get it now. Since we are changing the schema, you wonder if we should also track the schema version of conversations. I think is reasonable. Since it's named `schemaVersion` in the message model, what do you think of using the same name in the conversation model? Also, what about adding a metadata to the conversation schema?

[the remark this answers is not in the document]
aaronschachter: Yep - I think same name for both... I also wonder whether they should share the same schema version regardless if there's a change in one collection vs the other. I'm ok with adding `metadata` to conversation to keep consistent.

lib/helpers/request.js
```diff
@@ -383,6 +409,11 @@ function setMacro(req, macro) {
 function setTopic(req, topic) {
   req.topic = topic;
   analytics.addCustomAttributes({ topicId: topic.id });
+  /**
+   * FIXME: in the case of an askYesNo broadcast, because the topic here is the whole
+   * broadcast entry and the campaign info is inside the `saidYesTopic` nested entry.
+   * We do not set the campaign.
+   */
```
aaronschachter: I think the only reason we'd want to set the campaign is for checking to see whether the campaign has expired (we've traditionally always checked for the end date, but may not need to per https://www.pivotaltracker.com/story/show/164571771/comments/200498060)

We do have a check for when sending an outbound broadcast, we're checking by topic, not campaign, so seems like a non issue: https://github.com/DoSomething/gambit/blob/6.3.2/lib/middleware/messages/broadcast/broadcast-get.js#L18

We dont't have the check for it for users who may be responding to an `askYesNo` X days/months later, after the campaign has ended within the askYesNo middleware -- but even for this, we could check by `req.topic`  vs a `req.campaign`

app/models/Conversation.js
```diff
@@ -26,12 +27,17 @@ const conversationSchema = new mongoose.Schema({
...
+  /**
+   * Should we continue to set it? it wasn't being set correctly before PR #486 and
+   * broadcasts continued to work as expected.
```
aaronschachter: The bug here was that the `{{broadcast.id}}` tag would not be accurate.. and we do want to continue to set it to support that tag. Per https://dosomething.slack.com/archives/C1V0M6RPE/p1553112177006700?thread_ts=1553110258.004300&cid=C1V0M6RPE, I can verify by testing on `master`, the `{{broadcast.id}}` is not up to date when I say NO to an askYesNo that renders the tag (e.g. broadcast `4J5p6GrRUEzd9OjcwQvhHu`)

aaronschachter: This PR for sure fixes the bug...but where I get very 🤔 is setting a `topic.broadcastId` on the nested `broadcast.topic` --  as well as https://github.com/DoSomething/gambit/pull/486/files#r267612821. Wondering if https://github.com/DoSomething/gambit/pull/486/files#r267613663 is a way to potentially help make things a bit more readable/simple. For sure there's a bug in the `{{broadcast.id}}` tag but not necessarily needs to be part of the fix for saving the campaignId in the message of an outbound broadcast.

rapala61:
Closing so a fix to #488 can be included in a separate PR.

lib/helpers/request.js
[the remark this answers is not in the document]
aaronschachter: > This mess is likely better fixed with changes in GraphQL

I think it sounds sweet to add the `campaignId` property on the GraphQL side, handling the various checks based on broadcast type. Then on the Gambit side, we simply save the `campaignId` into the outbound message metadata if a `campaignId` exists in the response from getting/fetching the broadcast.

PR #465, opened 2019-01-24
Fetch broadcast from GraphQL

aaronschachter:
#### What's this PR do?

Continuing the deprecation of Gambit Content (#461, #463), this PR fetches and caches a broadcast via `broadcast(id)` GraphQL request instead of a `GET /broadcasts/:id request.`

#### How should this be reviewed?

Use Slack to test the following broadcasts:

* `7EqHlpg0Nv5czbOLYbbNPC` - askYesNo should work as expected

* `2bb1xED8Hi86OqWMYuQYO2` - photoPostBroadcast with attachment should work as expected

* `aWOjpNGcnXjm4GrpjI5uC` - autoReplyBroadcast should work as expected

* `33TAXhtxvqmSkk0AgqYYqg` - 422 error for askYesNo with saidYesTopic campaign ended

#### Any background context you want to provide?

Looking into how to fix the integration test that's failing... likely because we're now calling `getById` instead of `fetchById` in the get broadcast middleware of a `POST /messages` request with origins `broadcast`, `broadcastLite`.

#### Checklist
- [x] Tests added/updated for new features/bug fixes.
- [ ] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love.

PR #489, opened 2019-03-22
Fix conversation. lastReceivedBroadcastId

### LINKED ISSUE #488: {{broadcast.id}} renders outdated broadcast id.
_BUG_
-------------------
#### Current Behavior
- `conversation.lastReceivedBroadcastId` is not being updated as expected.

#### Desired Behavior
- `conversation.lastReceivedBroadcastId` should be updated with the last broadcast sent to the member. It's set so that when a user responds to an AskYesNo broadcast, the outbound message renders the broadcast id via the `{{broadcast.id}}` tag if present.

#### Steps to Replicate
- Locally, send yourself an AskYesNo broadcast that should send you a link to click on if you answer `Y` or `N`.  The un-shortened link should have an outdated broadcast id instead of the current broadcast.

#### Why This Matters
- Clicks tracked in Bertly would not be associated to the respective broadcast.

- [Slack conversation](https://dosomething.slack.com/archives/C2C8NLNAY/p1552409770058100)
- PR where the fix is discussed a bit more low level #486

aaronschachter:
#### What's this PR do?

Adds a `type` parameter to a broadcast to get the `isBroadcastable` topic helper working per #488. This bug was introduced in #465, when changing broadcast source content to GraphQL instead of Gambit Content.

Send yourself a broadcast, verify the conversation `lastReceivedBroadcastId` is set. Can also verify the `{{broadcast.id}}` tag renders correctly in NO responses to broadcasts like `1nnpJqvtHLJNbcF0GSVg5B`

Screen Shot 2019-03-22 at 7 01 58 AM

I've copied over the TODO of replacing the `type` property with `__typename`.. and eventually thought I'd take on those changes, but it touches quite a bit of code. I do think it's nice-to-have, as we can easily swap out the GraphQL types with different Contentful types.. but it'd most likely make sense to rename all of the helpers as well if we went this route (e.g. `isAskYesNoBroadcastTopic` vs `isAskYesNo`)

#### Relevant tickets
Fixes #488 

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/lib/helpers/broadcast.js
+++ b/lib/helpers/broadcast.js
@@ -15,6 +15,8 @@ async function fetchById(id) {
   logger.debug('Fetching broadcast', { id });
 
   const broadcast = await graphql.fetchBroadcastById(id);
+  // TODO: Refactor codebase to check for the topic GraphQL __typename, not the Contentful type name
+  Object.assign(broadcast, { type: broadcast.contentType });
 
   return helpers.cache.broadcasts.set(id, broadcast);
 }
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -27,10 +27,12 @@ const conversationSchema = new mongoose.Schema({
   },
   topic: String,
   campaignId: Number,
+  // Used to determine how to reply to an inbound messages for the conversation.
   lastOutboundMessage: {
     type: mongoose.Schema.Types.ObjectId,
     ref: 'Message',
   },
+  // Used to support rendering a {{broadcast}} tag in outbound messages for the conversation.
   lastReceivedBroadcastId: String,
 }, { timestamps: true });
 
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -26,7 +26,6 @@ const conversationSchema = new mongoose.Schema({
     index: true,
   },
   topic: String,
-  campaignId: Number,
   // Used to determine how to reply to an inbound messages for the conversation.
   lastOutboundMessage: {
     type: mongoose.Schema.Types.ObjectId,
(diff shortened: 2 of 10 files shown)


lib/helpers/broadcast.js
```diff
@@ -15,6 +15,8 @@ async function fetchById(id) {
   logger.debug('Fetching broadcast', { id });
 
   const broadcast = await graphql.fetchBroadcastById(id);
+  // TODO: Refactor codebase to check for the topic GraphQL __typename, not the Contentful type name
+  Object.assign(broadcast, { type: broadcast.contentType });
```
rapala61: Here is where we are going to start running into the inconsistencies I flagged in #486. This simple step does work for broadcasts that will be set as the topic since we would wait for the user's response to move them into a different topic. However, if we sent them a `photoPostPbroadcast`, this line will inject the `type`, inside the actual broadcast body. So when the `isBroadcastable` check [happens](https://github.com/DoSomething/gambit/blob/bug/broadcast-type/app/models/Conversation.js#L118), the topic won't show as broadcastable and the `lastReceivedBroadcastId` will not reflect this last broadcast.

aaronschachter: Ahh yes, that's a pickle and great flag. It seems rare any photo post or text post broadcasts would have change topics to templates that would contain a `{{broadcast.id}}` link, but could be very likely an auto reply broadcast would have. I also think this bug would have existed before GraphQL changes then.

Maybe the simplest fix could be to write the property somewhere in the [broadcast message middleware](https://github.com/DoSomething/gambit/blob/6.3.2/lib/middleware/messages/broadcast/conversation-update.js).... if we're sending a broadcast, update the property -- instead of trying to handle it when we're setting the conversation topic?

rapala61: > Maybe the simplest fix could be to write the property somewhere in the broadcast message middleware.... if we're sending a broadcast, update the property

This sounds very straight forward 💯 . If we set the `lastReceivedBroadcastId` every time we send a broadcast regardless of the type we would be effectively doing exactly what the name implies, beautiful.

aaronschachter: I've updated `lib/middleware/messages/broadcast/conversation-update` to fix saving `lastReceivedBroadcastId` per this approach. Doesn't seem like it hurts to leave setting the `type` property on a broadcast in this line to keep things consistent, but we don't use it anywhere because that helper checks for `contentType` -- and per these changes we don't need an `isBroadcastable` topic helper at all anymore.

PR #491, opened 2019-03-27
Change my name, change my name

aaronschachter:
#### What's this PR do?

Finds and replaces all instances of `Freddie` with `Tej`. I'll be adding a few comments per specific changes upon opening this PR.

#### How should this be reviewed?
* Text JOIN or LESS and confirm `Tej` is returned
* Text "Hi there" and confirm`Tej` is returned

Example:
```
You> hi there

Bot> Hi! This is Tej from DoSomething.org! I text about cool ways to take action in your community, AND scholarship opportunities. Go to DoSomething.org to see!
```

#### Any background context you want to provide?

Although we can override the welcome compliance text via config variable #487, it makes sense to change our hardcoded default, and merge/deploy when we're ready to make the switch.

#### Relevant tickets

https://dosomething.slack.com/archives/C2C8NLNAY/p1553707505112700?thread_ts=1553697026.111500&cid=C2C8NLNAY

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tested on staging.

PR #492, opened 2019-03-28
Brings back Ride and Seek Rivescript

aaronschachter:
#### What's this PR do?

Copies the Rivescript we used for Ride and Seek in #88 in preparation for its [return this May](https://dosomething.slack.com/archives/C2C8NLNAY/p1553699684111800?thread_ts=1553697026.111500&cid=C2C8NLNAY) (but changes out greetings per #491).

Sanity check the Rivescript by testing that the triggers send replies and change topics as expected per the [Rivescript syntax](https://www.rivescript.com/docs/tutorial#the-code-explained).

Yes! 

* We shouldn't merge this PR into `master` until we're ready to publish the keywords (e.g. `fomo`, `hero`) and go live with the game.

* These topics were originally named with the `seatbelt_` prefix because we didn't have a final campaign for a while, and it was a way to unblock hardcoding the replies. 

* The data team historically queried DB to find counts of messages within these topics to find out how many users completed the game, or got level to 5, etc -- which is why it'd be good to nail down what topic names we want to use this time around (whether to change or not)

### DIFF
--- a/brain/topics/seatbelt.rive
+++ b/brain/topics/seatbelt.rive
@@ -3,17 +3,57 @@
 /**
  * Game Start keywords
  */
-+ fomo
-- Hi, I’m Tej from DoSomething! I see our flyer caught your eye. So, "90% of this generation ___________ every weekend." Answer: wears a seat belt. Yep, almost *everyone* is doing it, but the 10% who aren’t are at serious risk.\s
-^ \n\nSeatbelts may seem kinda boring, but know what isn’t? *Literally* saving someone’s life. Want super, easy effective ways to keep friends safe? Text Y.{topic=seatbelt_level1}
++ flex
+- I see our flyer caught your eye. So, "What's stretchy, flexible and clicks?" Answer: A seat belt. In 2016, the use of seat belts in passenger vehicles saved an estimated 14,668 lives of occupants ages 5 and older. Let's keep saving lives by encouraging others to use their seat belts.\s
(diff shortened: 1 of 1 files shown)


ngjo:
Thanks for putting this together @aaronschachter ! Just confirming, this time around for Ride & Seek we'll also have keywords be hardcoded too or should Anthony be making these in Contentful?

aaronschachter:
For the smallest lift (we don't have to do anything besides edit this Rivescript and merge) the keywords will be hardcoded -- they're defined on line 6, 10, and 14 currently.

ngjo:
Perfect - I'll make sure that's included in the content when they're drafting it then.

brain/topics/seatbelt.rive
```diff
@@ -0,0 +1,176 @@
...
+
+< topic
+
+/**
+ * Final boss: Send NEXT
+ */
+> topic seatbelt_completed includes random
+
++ next
+- Now that you're a seat belt safety expert, want a fun and creative way to pass the knowledge on to your friends and family? Text SEEK to get started. {topic=random}
```
rapala61: This copy asks to text `SEEK` but I don't see the defaultTopicTrigger setup in Contentful. Is this on purpose for now? @DFurnes @ngjo

ngjo: I think it just hasn't been set up yet, but it's on Anthony's to do list!

PR #500, opened 2019-05-17
Adds broadcast id integration test

chloealee:
#### What's this PR do?
Adds a test to test the outbound broadcast message with {{broadcast.id}} tag should have the broadcast id rendered. 

Does this logic make sense to test this action? 

Creates test under "Outbound" section in the [Gambit Integration Tests](https://docs.google.com/document/d/1aBAMGIyKbYO7Ww6jAnnTSD2Fza1BMqnvplYNeqP4A48/edit?pli=1#heading=h.3hehsqy6rvtf) doc!

#### Relevant tickets
Adds first integration test to satisfy Pivotal [Card#165548479](https://www.pivotaltracker.com/n/projects/2328687/stories/165548479)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/test/integration/app/v2-routes/messages/broadcast.test.js
+++ b/test/integration/app/v2-routes/messages/broadcast.test.js
@@ -222,3 +222,34 @@ test('POST /api/v2/messages?origin=broadcast should save campaign id in outbound
   // TODO: DRY
   integrationHelper.hooks.cache.broadcasts.set(askYesNoBroadcast.id, null);
 });
+
+test('POST /api/v2/messages?origin=broadcast should save broadcast id in outbound data for ask yes no broadcasts', async (t) => {
+  const askYesNoBroadcast = stubs.graphql.getBroadcastSingleResponse('askYesNo').data.broadcast;
+
+  integrationHelper.hooks.cache.broadcasts.set(
+    askYesNoBroadcast.id,
+    askYesNoBroadcast,
+  );
+  const cioWebhookPayload = stubs.broadcast.getCioWebhookPayload(askYesNoBroadcast.id);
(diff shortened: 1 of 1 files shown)


test/integration/app/v2-routes/messages/broadcast.test.js
```diff
@@ -222,3 +222,34 @@ test('POST /api/v2/messages?origin=broadcast should save campaign id in outbound
...
+    }))
+    .set('Authorization', `Basic ${integrationHelper.getAuthKey()}`)
+    .send(cioWebhookPayload);
+
+  res.status.should.be.equal(200);
+  res.body.data.messages.length.should.be.equal(1);
+  res.body.data.messages[0].broadcastId.should.equal(cioWebhookPayload.broadcastId);
+
+  // clear cache
+  // TODO: DRY
```
chloealee: @rapala61 I added these since I followed the test above as an example to create this test. Why should we need to clear the cache because I thought that is what `test.beforeEach` does in the beginning of this file? Also not sure exactly what we should DRY up!

rapala61: The test looks good! As far as the cache clearing, we should just empty the cache completely through a function call in `test.afterEach` instead of at the end of each test. That is that the "DRYing" comment is about. Better yet, instead of relaying on the hack of warming up the cache, we should use an interceptor for that! Updating that code is out of the scope of the test challenge so it's fine to leave for a later PR.

chloealee: ok thank you for the explanation for that - makes total sense! Thanks, @rapala61 !

PR #501, opened 2019-05-20
Adds compliance text tests

chloealee:
#### What's this PR do?
Adds test to make sure user receives compliance text (outbound reply has template `sendInfoMessage`) if they send INFO or HELP. Also, tests that their `topic` does not change.

Do the tests pass? 
Creates "User received compliance text if they send INFO or HELP" test in the 
[Gambit Integration Tests](https://docs.google.com/document/d/1aBAMGIyKbYO7Ww6jAnnTSD2Fza1BMqnvplYNeqP4A48/edit?pli=1#heading=h.3hehsqy6rvtf) doc.

#### Relevant tickets
Adds second integration test to satisfy Pivotal [Card#165548479](https://www.pivotaltracker.com/n/projects/2328687/stories/165548479)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "gambit",
-  "version": "6.7.2",
+  "version": "6.7.3",
   "description": "The DoSomething.org chatbot service for multi-platform conversations.",
   "author": "DoSomething.org",
   "license": "MIT",
--- a/test/helpers/integration/routes.js
+++ b/test/helpers/integration/routes.js
@@ -5,6 +5,8 @@ const querystring = require('querystring');
 
 const northstarConfig = require('../../../config/lib/northstar');
 const graphqlConfig = require('../../../config/lib/graphql');
+const bertlyConfig = require('../../../config/lib/bertly');
+
 
 /**
  * TODO: Use the Node.js url module to standardize the return of all the methods
@@ -68,8 +70,9 @@ module.exports = {
     },
   },
   bertly: {
+    baseURI: bertlyConfig.baseURI,
     intercept: function intercept(reply = {}, times = 1, status = 200) {
-      return nock('http://shorturl')
+      return nock(this.baseURI)
         .post('/')
         .times(times)
         .reply(status, reply);
--- a/test/helpers/integration/routes.js
+++ b/test/helpers/integration/routes.js
@@ -71,9 +71,9 @@ module.exports = {
   },
   bertly: {
     baseURI: bertlyConfig.baseUri,
-    intercept: function intercept(reply = {}, times = 1, status = 200) {
+    intercept: function intercept(urlPath, reply = {}, times = 1, status = 200) {
       return nock(this.baseURI)
-        .post('/')
+        .post(`/${urlPath}`)
         .times(times)
         .reply(status, reply);
     },
--- a/test/helpers/integration/routes.js
+++ b/test/helpers/integration/routes.js
@@ -67,4 +67,13 @@ module.exports = {
         .reply(status, reply);
     },
   },
+  bertly: {
+    baseURI: graphqlConfig.clientOptions.baseURI,
+    intercept: function intercept(reply = {}, times = 1, status = 200) {
+      return nock(this.baseURI)
+        .post('http://shorturl')
+        .times(times)
+        .reply(status, reply);
+    },
+  },
 };
--- a/test/helpers/integration/routes.js
+++ b/test/helpers/integration/routes.js
@@ -70,7 +70,7 @@ module.exports = {
   bertly: {
     intercept: function intercept(reply = {}, times = 1, status = 200) {
       return nock('http://shorturl')
-        .post()
+        .post('/')
         .times(times)
         .reply(status, reply);
     },
--- a/test/helpers/integration/routes.js
+++ b/test/helpers/integration/routes.js
@@ -71,7 +71,7 @@ module.exports = {
   },
   bertly: {
     baseURI: bertlyConfig.baseUri,
-    intercept: function intercept(urlPath, reply = {}, times = 1, status = 200) {
+    intercept: function intercept(reply = {}, times = 1, status = 200) {
       return nock(this.baseURI)
         .post()
         .times(times)
--- a/test/helpers/integration/routes.js
+++ b/test/helpers/integration/routes.js
@@ -68,10 +68,9 @@ module.exports = {
     },
   },
   bertly: {
-    baseURI: graphqlConfig.clientOptions.baseURI,
     intercept: function intercept(reply = {}, times = 1, status = 200) {
-      return nock(this.baseURI)
-        .post('http://shorturl')
+      return nock('http://shorturl')
+        .post()
         .times(times)
         .reply(status, reply);
     },
--- a/test/helpers/integration/routes.js
+++ b/test/helpers/integration/routes.js
@@ -73,7 +73,7 @@ module.exports = {
     baseURI: bertlyConfig.baseUri,
     intercept: function intercept(urlPath, reply = {}, times = 1, status = 200) {
       return nock(this.baseURI)
-        .post(`/${urlPath}`)
+        .post()
         .times(times)
         .reply(status, reply);
     },
--- a/test/helpers/integration/routes.js
+++ b/test/helpers/integration/routes.js
@@ -70,7 +70,7 @@ module.exports = {
     },
   },
   bertly: {
-    baseURI: bertlyConfig.baseURI,
+    baseURI: bertlyConfig.baseUri,
     intercept: function intercept(reply = {}, times = 1, status = 200) {
       return nock(this.baseURI)
         .post('/')
--- a/test/integration/app/v2-routes/messages/broadcast.test.js
+++ b/test/integration/app/v2-routes/messages/broadcast.test.js
@@ -258,7 +258,7 @@ test('POST /api/v2/messages?origin=twilio outbound message should match sendInfo
   const message = {
     getSmsMessageSid: stubs.twilio.getSmsMessageSid(),
     From: stubs.getMobileNumber('valid'),
-    Body: "info",
+    Body: 'info',
   };
 
   const res = await t.context.request
@@ -276,7 +276,7 @@ test('POST /api/v2/messages?origin=twilio outbound message should match sendInfo
   const message = {
     getSmsMessageSid: stubs.twilio.getSmsMessageSid(),
     From: stubs.getMobileNumber('valid'),
-    Body: "help",
+    Body: 'help',
   };
 
   const res = await t.context.request
--- a/test/integration/app/v2-routes/messages/broadcast.test.js
+++ b/test/integration/app/v2-routes/messages/broadcast.test.js
@@ -254,39 +254,3 @@ test('POST /api/v2/messages?origin=broadcast should save broadcast id in outboun
   integrationHelper.hooks.cache.broadcasts.set(cioWebhookPayload.broadcastId, null);
 });
 
-test('POST /api/v2/messages?origin=twilio outbound message should match sendInfoMessage if user texts INFO', async (t) => {
-  const message = {
-    getSmsMessageSid: stubs.twilio.getSmsMessageSid(),
-    From: stubs.getMobileNumber('valid'),
-    Body: 'info',
-  };
-
-  const res = await t.context.request
-    .post(integrationHelper.routes.v2.messages(false, {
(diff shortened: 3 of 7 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "gambit",
-  "version": "6.7.2",
+  "version": "6.7.3",
```
chloealee: hm for some reason this didn't update when I deployed #501 to prod 🤔

chloealee:
hmm @rapala61 looks like I'm getting a `403` status code for these tests on wercker (but all is fine on my local!). I don't think that it has anything to do with the number that the text is from because anyone should be able to hit this endpoint without any permissions specified?

test/integration/app/v2-routes/messages/broadcast.test.js
```diff
@@ -253,3 +253,40 @@ test('POST /api/v2/messages?origin=broadcast should save broadcast id in outboun
   // TODO: DRY
   integrationHelper.hooks.cache.broadcasts.set(cioWebhookPayload.broadcastId, null);
 });
+
+test('POST /api/v2/messages?origin=twilio outbound message should match sendInfoMessage if user texts INFO', async (t) => {
```
rapala61: These tests should be moved to the [`member.test.js`](https://github.com/DoSomething/gambit/blob/complience-text-test/test/integration/app/v2-routes/messages/member.test.js) file.

test/integration/app/v2-routes/messages/broadcast.test.js
```diff
@@ -253,3 +253,40 @@ test('POST /api/v2/messages?origin=broadcast should save broadcast id in outboun
...
+  const message = {
+    getSmsMessageSid: stubs.twilio.getSmsMessageSid(),
+    From: stubs.getMobileNumber('valid'),
+    Body: 'info',
+  };
+
+  const res = await t.context.request
```
rapala61: This http request will trigger a 3rd party call to Northstar to find the user texting `HELP` or `INFO`. Same with the second test below. For this reason, we need to intercept the call and return a mocked user. An example of how that works is [here](https://github.com/DoSomething/gambit/blob/complience-text-test/test/integration/app/v2-routes/messages/broadcast.test.js#L238-L239).

The reason these tests are working locally is because you have correct `DS_NORTHSTAR_API_OAUTH_* `credentials in your local `.env` file. Since we are not intercepting the call, they are actually reaching NS when you test locally. They are not setup in wercker so the tests are failing there.

test/integration/app/v2-routes/messages/broadcast.test.js
chloealee: ah yes - moved!

test/integration/app/v2-routes/messages/broadcast.test.js
chloealee: ah this makes a lot of sense, thanks for explaining, @rapala61 ! 

I updated these two tests and they aren't passing locally, returning a `404` error. I think that this is because it can't find the user in [this middleware](https://github.com/DoSomething/gambit/blob/master/lib/middleware/messages/user-get.js#L20). This makes sense too because I searched aurora thor for [this northstar_id](https://github.com/DoSomething/gambit/blob/master/test/helpers/stubs.js#L324) and it doesn't exist.

However, I'm not sure how this is working [here](https://github.com/DoSomething/gambit/blob/complience-text-test/test/integration/app/v2-routes/messages/broadcast.test.js#L238-L239)? `cioWebhookPayload.userId` also uses this same [`getUserId()`](https://github.com/DoSomething/gambit/blob/master/test/helpers/stubs.js#L216) function so it should be the same user id being used.

chloealee: @rapala61 ah I think I got it and now just failing because the link in the message is incorrect!
```
AssertionError {
    actual: 'These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or http://doso.me/1jf4/291kep. Txt STOP to quit. Msg&Data Rates May Apply.',
    expected: 'These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or https://dosome.click/zt6mc. Txt STOP to quit. Msg&Data Rates May Apply.',
    message: 'expected \'These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or http://doso.me/1jf4/291kep. Txt STOP to quit. Msg&Data Rates May Apply.\' to equal \'These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or https://dosome.click/zt6mc. Txt STOP to quit. Msg&Data Rates May Apply.\'',
    showDiff: true,
  }
```

Is this link different on QA for some reason? I took this text from [the helper](https://github.com/DoSomething/gambit/blob/9c3560f4bb64462da661aa9891c3e0584263b7fc/config/lib/helpers/macro.js#L165) so assuming this should be the same everywhere?

rapala61: I think what is going on is that the embedded link in the text is being re-shortened by Bertly and that is why it won't match. The way to go here is to intercept the Bertly link and mock it with any url that you can retrieve in the test so you can compare.

rapala61: Here is where we shorten links in outbound texts: https://github.com/DoSomething/gambit/blob/complience-text-test/app/models/Conversation.js#L245.

test/integration/app/v2-routes/messages/member.test.js
```diff
@@ -104,3 +104,73 @@ test.serial('POST /api/v2/messages?origin=twilio should not re-send message to T
...
+  // mock user fetch
+  integrationHelper.routes.northstar
+    .intercept.fetchUserByMobile(message.From, member, 1);
+
+  // mock user update
+  integrationHelper.routes.northstar
+    .intercept.updateUserById(member.data.id, member, 1);
+
+  integrationHelper.routes.northstar
+    .intercept.fetchUserById(stubs.getUserId(), member, 1);
```
rapala61: I think code from line `114` to `130` could be moved into a function at the top of this test file and used in the next test too.

test/integration/app/v2-routes/messages/member.test.js
```diff
@@ -104,3 +104,73 @@ test.serial('POST /api/v2/messages?origin=twilio should not re-send message to T
...
+
+  const res = await t.context.request
+    .post(integrationHelper.routes.v2.messages(false, {
+      origin: 'twilio',
+    }))
+    .set('Authorization', `Basic ${integrationHelper.getAuthKey()}`)
+    .send(message);
+  res.status.should.be.equal(200);
+  res.body.data.messages.outbound[0].text.should.equal('These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or https://dosome.click/zt6mc. Txt STOP to quit. Msg&Data Rates May Apply.');
+  res.body.data.messages.inbound[0].topic.should.equal(res.body.data.messages.outbound[0].topic);
```
rapala61: This assertion is really not too helpful because we don't have a trusted expected value to compare against. Here we are just comparing the "actual" topic to itself. What we need is to find where the topic is in a configuration file and compare against that value.

chloealee: ah ok thanks @rapala61 ! I wrote this test this way to make sure the inbound and outbound topics didn't change to test this (in case it changes topic in the future) but not sure if is correct since you're right, the inbound topic could be incorrect as well. Just updated this to make sure the topic is still "random"!

test/integration/app/v2-routes/messages/broadcast.test.js
chloealee: ooo ok that makes sense, thanks @rapala61 ! To intercept that Bertly call, would I want to build something out like [this snippet](https://github.com/DoSomething/gambit/blob/master/test/helpers/integration/routes.js#L61-L69)?

rapala61: Yes! you can follow the pattern there 👍.

chloealee: awesome, thanks @rapala61 ! Just mocked it out and ready for another review!

chloealee: spoke too soon - still not passing the wercker tests but I'll give a shout when these are passing!

chloealee: hi @rapala61 ! It seems like my changes broke [this test](https://github.com/DoSomething/gambit/blob/beecf5a28f61aa25fce5e196c4b5ba8e5ec072b2/test/unit/lib/consolebot.test.js#L104) but I'm not sure why since the changes are only in the `member.test.js` file to mock bertly calls and seems like the test that broke is in `consolebot.test.js` file? 

Screen Shot 2019-06-03 at 10 53 42 AM

rapala61: Taking a look now!

test/helpers/integration/routes.js
```diff
@@ -67,4 +67,12 @@ module.exports = {
         .reply(status, reply);
     },
   },
+  bertly: {
+    intercept: function intercept(reply = {}, times = 1, status = 200) {
+      return nock('http://shorturl')
```
rapala61: We should be definitely using the Bertly URL setup in the env variable just like we do for [GraphQL](https://github.com/DoSomething/gambit/blob/complience-text-test/test/helpers/integration/routes.js#L62). You can then use "http://shorturl" as the value for the `DS_BERTLY_API_BASEURI` ENV variable in your local `.env` file -- I already added it to the Wercker's env variables.

test/helpers/integration/routes.js
```diff
@@ -67,4 +69,13 @@ module.exports = {
         .reply(status, reply);
     },
   },
+  bertly: {
+    baseURI: bertlyConfig.baseURI,
```
rapala61: Small typo, should be `bertlyConfig.baseUri`. The naming could be more consistent so it doesn't promote these errors.

chloealee: ah yes, thanks for catching that, @rapala61 !

chloealee:
@rapala61 hm from the wercker build and on my local, it seems like it is now not mocking the bertly call? 

```
  AssertionError {
    actual: 'These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or http://doso.me/1jf4/291kep. Txt STOP to quit. Msg&Data Rates May Apply.',
    expected: 'These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or https://dosome.click/zt6mc. Txt STOP to quit. Msg&Data Rates May Apply.',
    message: 'expected \'These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or http://doso.me/1jf4/291kep. Txt STOP to quit. Msg&Data Rates May Apply.\' to equal \'These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or https://dosome.click/zt6mc. Txt STOP to quit. Msg&Data Rates May Apply.\'',
    showDiff: true,
  }
```

test/integration/app/v2-routes/messages/member.test.js
```diff
@@ -104,3 +127,43 @@ test.serial('POST /api/v2/messages?origin=twilio should not re-send message to T
...
+  mockExternalCallsForUserInputMessage(message);
+
+  const res = await t.context.request
+    .post(integrationHelper.routes.v2.messages(false, {
+      origin: 'twilio',
+    }))
+    .set('Authorization', `Basic ${integrationHelper.getAuthKey()}`)
+    .send(message);
+  res.status.should.be.equal(200);
+  res.body.data.messages.outbound[0].text.should.equal('These are Do Something Alerts - 4 messages/mo. Info help@dosomething.org or https://dosome.click/zt6mc. Txt STOP to quit. Msg&Data Rates May Apply.');
```
rapala61: @chloealee I was able to make this test pass by replacing this hardcoded string w/ the text in the [macro helper](https://github.com/DoSomething/gambit/blob/complience-text-test/config/lib/helpers/macro.js#L165) for `sendInfoMessage` Rivescript macro (What is sent to the user when they text INFO or HELP).

You will want to require the config in this test file and reference the text of the macro instead of the hardcoded string in both tests.

chloealee: yay thank you so much @rapala61 ! This is all passing now and ready for final review!

chloealee:
Thanks for all of your help with this, @rapala61 !!

PR #526, opened 2020-09-02
Updates Gambit workflow documentation

aaronschachter:
#### What's this PR do?

This PR documents that we no longer will use Wunderflow to maintain a `production` branch with tags. Instead, we'll maintain a `main` branch (which is our `master` branch renamed), which our staging app will automatically build from -- and manually promote to production for deploys.

 I've made these changes to the Heroku pipeline, and this PR seemed like a good test run for a deploy (it's been almost a year!)

👀 

I'm reading this PR template for the first time in ages, and remembering that we'd manually deploy branches to our staging app to test them out. Review apps are tough to do in this project because we'd need to change the URL's for the various front-ends (Slack app, Twilio phone number). We've usually been able to get away with using our staging app as a review app because there's rarely other folks using it at the same time.

It does bring to mind a wishlist item of moving Gambit Admin into this repo, and building out a Consolebot so devs / stakeholders could test out conversations in the UI of the review app (but even that'd require adding the review app into the dev Aurora instance like we need to for Phoenix - and we'd need to add it each time since we wouldn't want to expose the chat widget to anonymous users)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/.github/README.md
+++ b/.github/README.md
@@ -74,7 +74,6 @@ You>
 
 Input is posted to your localhost `api/v2/messages?origin=twilio` endpoint on behalf of the Northstar User with mobile number matching your  `DS_CONSOLEBOT_USER_MOBILE` config variable. A new Northstar User is created for the mobile number if it doesn't exist.
 
-
 ### Contributing
 
 * Run `npm test:full` to lint code and run automated tests.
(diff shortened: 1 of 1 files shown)


katiecrane (approved):
love the consistency this will keep with our other apps!

PR #527, opened 2020-09-03
Adds support for LocationVotingInformation tags

aaronschachter:
#### What's this PR do?

This PR adds support to query GraphQL for the `LocationVotingInformation` relevant to a user per their `addr_state` Northstar property, and display them as new [user tags](https://github.com/dosomething/gambit-admin/wiki/tags#user).

Example message content:

> Don't forget! Early voting in {{user.addrState}} beings on {{user.earlyVotingStarts}} and ends on {{user.earlyVotingEnds}}.

Example rendered message:

> Don't forget! Early voting in MA beings on 9/12 and ends on 10/13.

This can be manually tested by sending yourself this [test broadcast, 7BKADblVGm0o67HyORvDm4,](https://app.contentful.com/spaces/owik07lyerdj/entries/7BKADblVGm0o67HyORvDm4) which includes `earlyVotingStarts` and `earlyVotingEnds` user tags.

* A few functions and related tests needed to be refactored as async to do this, and I'm second guessing making the GraphQL query from the tags helper, vs adding into get/create user functions -- but because there are so many routes that load a user, this still feels like maybe the cleanest. 

* Adding Prettier to this repo's commit hooks would truly be the jam

* I'll update the Gambit wiki to list these new [tags](https://github.com/dosomething/gambit-admin/wiki/tags#user), and mull over where the ideal home of SMS documentation for non-devs should live (for now, it's here and I'd think Anthony's google doc)

#### Relevant tickets

* Refs https://github.com/DoSomething/graphql/pull/275

* Closes https://www.pivotaltracker.com/story/show/174179498

#### Checklist
- [x] Tested on staging.

### DIFF
--- a/app/models/Conversation.js
+++ b/app/models/Conversation.js
@@ -286,9 +286,11 @@ conversationSchema.methods.getMessagePayloadFromReq = function (req = {}, direct
  */
 conversationSchema.methods.createMessage = async function (direction, text, template, req) {
   logger.debug('createMessage', { direction }, req);
+
   let messageText;
+
   if (direction !== 'inbound') {
-    messageText = helpers.tags.render(text, req);
+    messageText = await helpers.tags.render(text, req);
 
     if (bertly.isEnabled() && bertly.textHasLinks(messageText)) {
       messageText = await bertly.parseLinksIntoRedirects(messageText);
--- a/config/lib/graphql.js
+++ b/config/lib/graphql.js
@@ -250,11 +250,25 @@ const fetchWebSignupConfirmations = `
   ${campaignTopicTransitionFragments}
 `;
 
+const fetchVotingInformationByLocation = `
+  query getLocationVotingInformation($location: String!) {
+    locationVotingInformation(location: $location) {
+      absenteeBallotRequestDeadline
+      absenteeBallotReturnDeadline
+      absenteeBallotReturnDeadlineType
+      earlyVotingEnds
+      earlyVotingStarts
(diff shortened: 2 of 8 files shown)


aaronschachter:
I'm going to deploy this to our staging app to do a little manual testing.

aaronschachter:
Confirmed the test broadcast renders the tags, and that replies and signup confirmations are working as expected.

## Broadcast / replies

Screen Shot 2020-09-03 at 1 55 21 PM

## Signup

Screen Shot 2020-09-03 at 1 57 55 PM

mendelB (approved):
I'm still a major newb around here, but this looks sound to me! 👍 

> I'm second guessing making the GraphQL query from the tags helper, vs adding into get/create user functions -- but because there are so many routes that load a user, this still feels like maybe the cleanest.

I can't speak much to this, but it _seems_ like it could be nice to at least somehow isolate the voter information tag logic A. since it has it's own unique functionality and B. so that we're only querying GraphQL for voter reg broadcasts or messages specifically (although I suppose we don't differentiate this and these are simply additional supported tags, so probably not too feasible at this point?) (but also I don't really know wtf I'm talking about)

aaronschachter:
Yeah, I think for now this is the easiest way to ship -- which is querying GraphQL for every outbound message, even if we don't have any voting related tags within the message text. A follow up PR might be to first check the message text for whether there are instances of open/closed double parenthesis, and then call render.. but for now let's go with this as a first pass.

PR #529, opened 2020-09-04
Avoid querying GraphQL if outbound message does not have tags

aaronschachter:
#### What's this PR do?

This PR adds a check into the tags helper's `render` function so that we only call `getTags` (which queries GraphQL for a user's `LocationVotingInformation`) if opening and closing double brackets exist within the outbound message text. Refs https://github.com/DoSomething/gambit/pull/527#issuecomment-686774337

👀 

I've been having a tough time figuring out how to run individual unit tests instead of `npm run test:unit`, which runs them all. I've been trying to use [`npx ava --match='*test*'`](https://github.com/avajs/ava/blob/master/docs/05-command-line.md#running-tests-with-matching-titles), but some tests fail (whereas they won't when running all tests via the `npm run test:unit` command). 

#### Relevant tickets

[#174179498](https://www.pivotaltracker.com/story/show/174179498)

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/helpers/tags.js
+++ b/lib/helpers/tags.js
@@ -3,6 +3,7 @@
 const mustache = require('mustache');
 const queryString = require('query-string');
 
+const logger = require('../logger');
 const graphql = require('../graphql');
 const config = require('../../config/lib/helpers/tags');
 const userConfig = require('../../config/lib/helpers/user');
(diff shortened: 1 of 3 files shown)


lib/helpers/tags.js
```diff
@@ -160,12 +165,17 @@ function getLinksTag(req) {
...
+ * Renders Mustache tags with req data, if tags are included in the given string.
+ *
  * @param {String} string
  * @param {Object} req
  * @return {String}
  */
 async function render(string, req) {
-  return mustache.render(string, await module.exports.getTags(req));
+  // Check for opening and closing double brackets.
+  const hasTags = /\{{.*?\}}/g.exec(string);
```
mendelB, line 176: Could this use the one-or-more `+?`
```suggestion
  const hasTags = /\{{.+?\}}/g.exec(string);
```

So we don't match on blank tags? (edge case-y - yes).

mendelB (approved):
👏

PR #532, opened 2020-09-30
Update SMS Voice & Compliance/Welcome Copy

mendelB:
#### What's this PR do?
tell us, what have you changed?
- We've updated our compliance copy / welcome message to introduce the new cohort of SMS voices
- We've updated the official Rivescript bot `name` from `Tej` to `Ben` along with any hardcoded references

tell us, how can we review, to test that this works
Am I missing any other references? I noticed some documentation & stubs still referencing Tej, but these seemed to clearly reference historical or example messaging, so wasn't sure if it was necessary to update.

what else?
With Tej's departure ( 😢 ), we've migrated to a cohort of SMS voices, with Ben being the de facto bot `name`. ([See Slack](https://dosomething.slack.com/archives/C03T8SDMS/p1601484313001400?thread_ts=1601415311.001100&cid=C03T8SDMS))

#### Relevant tickets
Fixes #
Fixes Pivotal [Card#175060670](https://www.pivotaltracker.com/story/show/175060670)

#### Checklist
- [ ] Tested on staging.

image

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -58,7 +58,7 @@ DS_CONSOLEBOT_USER_MEDIA_URL=http://placekitten.com/g/400/400
 DS_CONSOLEBOT_USER_MEDIA_TYPE=image/jpg
 
 DS_GAMBIT_CONVERSATIONS_COMPLETED_VOTING_PLAN_TEXT=Great job!
-DS_GAMBIT_CONVERSATIONS_SUBSCRIPTION_STATUS_ACTIVE_TEXT="Hi I'm Tej from DoSomething.org! Welcome to my weekly updates (up to 8msg/month). Things to know: Msg&DataRatesApply. Text HELP for help, text STOP to stop."
+DS_GAMBIT_CONVERSATIONS_SUBSCRIPTION_STATUS_ACTIVE_TEXT="Hi I'm Ben from DoSomething.org! Welcome to my weekly updates (up to 8msg/month). Things to know: Msg&DataRatesApply. Text HELP for help, text STOP to stop."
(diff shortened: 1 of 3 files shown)


mendelB:
As an aside @aaronschachter, there didn't seem to be many places with a hardcoded name (aside from the two updates in the `macro.js`), so this could just be overkill -- but wondering if we could process those messages through the Rivescript templating similar to Contentful content (I'm speaking roughly because I'm not familiar with the nuances), so that instead of hardcoding a name we could use the `<bot name>` templating?

aaronschachter:
It'd be sweet to use`<bot name>` instead of hardcoding the `Ben`, but this message text is set after we've already sent the inbound message to the Rivescript middleware (and got the macro name back, like `subscriptionStatusActive` or `subscriptionStatusLess`):
https://github.com/DoSomething/gambit/blob/a3ebe2c96c32ff8b3e42453352562ed58742d8cb/brain/macros.rive#L15-L25

It could be nice to potentially call a Rivescript function to replace any `<bot>` variables in macro text... but don't think we need to worry about it for this pass.

aaronschachter (approved):
> Am I missing any other references?

I don't think so!

mendelB:
Ah thanks for that context @aaronschachter and agreed for now it certainly doesn't seem like we need to worry about this. Especially because I updated in https://github.com/DoSomething/gambit/pull/532/commits/7a109569cbc6b0f51f58fd7939f7da95eaf40abd to remove the remaining hardcoded name entirely per [Anthony in Slack](https://dosomething.slack.com/archives/C03T8SDMS/p1601490077003500?thread_ts=1601415311.001100&cid=C03T8SDMS)!

PR #545, opened 2021-01-20
Switch from Wercker to CircleCI

aaronschachter:
#### What's this PR do?

This PR configures this repo to use CircleCI instead of Wercker to maintain parity with the majority of our other apps.

👀 

* One gotcha I caught in the process is that some tests querying GraphQL directly. I fixed a unit test by mocking its GraphQL request, but there are a few integration tests that broke until I added the `DS_GRAPHQL_API_BASEURI` config variable -- suggesting there are more network requests that should be mocked... 

* I didn't add codecov into the CircleCI steps for now, can potentially add that back in later (need to dig around to see if I have a login...)

#### Relevant tickets

https://www.pivotaltracker.com/story/show/176564578

#### Checklist
- [ ] Tested on staging.


.circleci/config.yml
```diff
@@ -0,0 +1,15 @@
...
+    docker:
+      - image: circleci/node:10.16.3
+      - image: circleci/mongo:3.6.21
+      - image: circleci/redis
+    steps:
+      - checkout
+      - run: npm install
+      - run: npm run lint
+      - run: npm run test:unit
+      - run: npm run test:integration
```
mendelB, line 15: Are we using these individually to avoid the `npm run coverage` in the `npm run test:full` command?

mendelB (approved):
🟢

.circleci/config.yml
aaronschachter: Yeah, I'm holding off on adding codecov in for now. it seems CircleCI doesn't need a token from Codecov, which we used for Werkcer.

DFurnes (commented):
Heck yeah! 🚥

PR #547, opened 2021-01-25
Updates Node to version 12, npm updates

aaronschachter:
### What's this PR do?

This pull request updates the Node version to 12 and commits changes per `npm update.`

### How should this be reviewed?

👀 

### Any background context you want to provide?

Going to look into updating our Mongoose package to v5 in a separate PR per needing to check with the [migration guide](https://mongoosejs.com/docs/migrating_to_5.html).

### Relevant tickets

References [Pivotal #176647327](https://www.pivotaltracker.com/story/show/176647327).

#### Checklist
- [ ] Documentation added for new features/changed endpoints.
- [ ] Tests added/updated for new features/bug fixes.
- [ ] ENV variables added/updated for new features/bug fixes.
- [ ] Tested on staging.
- [ ] Pinged a PM if this is a larger PR that would benefit from some additional testing love.

PR #549, opened 2021-01-25
Removes npm key from engines

aaronschachter:
### What's this PR do?

This pull request follows instructions in [this Heroku article](https://help.heroku.com/ZV7S7D6T/why-is-my-node-build-is-suddenly-displaying-npm-err-cb-apply-is-not-a-function) to avoid the ` npm ERR! cb.apply is not a function` error that occurred when trying to build #547 

👀 

🚑 

### Relevant tickets

References [Pivotal #176647327](https://www.pivotaltracker.com/story/show/176647327).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,7 @@
     ]
   },
   "engines": {
-    "node": "12.20.1",
-    "npm": "6.9.0"
+    "node": "12.20.1"
   },
   "dependencies": {
     "@dosomething/gateway": "^2.0.1",
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     ]
   },
   "engines": {
-    "node": "12.20.1"
+    "node": "12.x"
   },
   "dependencies": {
     "@dosomething/gateway": "^2.0.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -38,8 +38,7 @@
     ]
   },
   "engines": {
-    "node": "12.20.1",
-    "npm": "6.9.0"
+    "node": "12.20.1"
```
DFurnes, line 41: It looks like Heroku recommends [only specifying the major version number](https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version), like `12.x`, so we get security fixes.

DFurnes (approved):
Just one unrelated suggestion! 🔑

package.json
aaronschachter: Good call.It seems that for the CircleCi config, [we'd still need to specify 12.20](https://circleci.com/developer/images/image/cimg/node) (not just 12). I'm thinking of upgrading to 14 next, so we can use the `node:lts` there -- but I wanted to upgrade incrementally 👶 👣

PR #551, opened 2021-02-02
Inspect action.campaign for photo/text post topics

aaronschachter:
### What's this PR do?

This pull request modifies the request helper to first check for whether a topic has an action set per new changes in https://github.com/DoSomething/graphql/pull/320.

If the topic does have an action set, use the nested action.campaign to set the `req.campaign` for the request. We'll eventually be removing the `campaign` reference field from the photo/text post topic content types, as it's redundant.

👀 

Once this is in place, we can remove the `campaign` reference field from the `TextPostTopic` and `PhotoPostTopic` GraphQL types, and delete the reference field in Contentful.

### Relevant tickets

References [Pivotal #176602032](https://www.pivotaltracker.com/story/show/176602032).

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/helpers/request.js
+++ b/lib/helpers/request.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const lodash = require('lodash');
+
 const helpers = require('../helpers');
 const logger = require('../logger');
 const analytics = require('./analytics');
@@ -324,10 +326,7 @@ function saveDraftSubmissionValue(req, key, value) {
 function setCampaign(req, campaign) {
   req.campaign = campaign;
   analytics.addCustomAttributes({ campaign });
-  /**
-   * TODO: If we remove the campaign field in contentful for some topics. We will have to
-   * tell GraphQL to fetch it from Rogue based on the action Id?
-   */
+
   if (campaign && campaign.id) {
     module.exports.setCampaignId(req, campaign.id);
   }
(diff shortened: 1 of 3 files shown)


lib/helpers/request.js
```diff
@@ -383,9 +382,11 @@ function setMacro(req, macro) {
 function setTopic(req, topic) {
   req.topic = topic;
   analytics.addCustomAttributes({ topicId: topic.id });
-  if (topic.campaign) {
-    module.exports.setCampaign(req, topic.campaign);
-  }
+
+  module.exports.setCampaign(
+    req,
+    lodash.get(topic, 'action.campaign', null) || topic.campaign,
```
mendelB, line 388: When would the `topic.campaign` be populated if we removed that field from the GraphQL query above?

mendelB (approved):
👍

lib/helpers/request.js
aaronschachter: An `AutoReplyTopic` is the one instance where a topic could have a `campaign` set without an `action`.

aaronschachter:
This work would require a bit more refactoring than I anticipated -- we'd need to make similar changes to the `executeInboundTopicChange` request helper, as well as the `hasActiveCampaign` topic helper it calls:

https://github.com/DoSomething/gambit/blob/700b46e3fcbe98327384532bea8809743c4b7d7e/lib/helpers/request.js#L43-L48

Holding off on this for now,  as I thought I'd be able to knock this out today -- but we'll need to save it for a future cleanup sprint.

PR #552, opened 2021-03-16
Adds broadcastLite route for Northstarless broadcasts

aaronschachter:
### What's this PR do?

This pull request resurrects the `broadcastLite` origin parameter for a `POST /v2/messages` API request to send a Northstarless broadcast,  excluding the `getUser` middleware that queries Northstar to validate SMS status.

When the `broadcastLite` origin is passed instead of `broadcast`, we require all user properties needed for sending a broadcast to be sent as API request parameters -- which we can easily do from Customer.io by adding each user's profile property -- and use to them set the `req.user` that the other middleware depend on.

Example request:
```
curl --location --request POST 'http://localhost:5100/api/v2/messages?origin=broadcastLite' \
--header 'Content-Type: application/json' \
--header 'Authorization: Basic cHVwcGV0OnRvdGFsbHlzZWNyZXQ=' \
--data-raw '{
    "userId": "5547be89469c64ec7d8b518d",
    "broadcastId": "2IkRmKYUqySjPTEEHDS8q1",
    "mobile": "[removed]",
    "smsStatus": "active"
}'
```

Example response:
```
{
    "data": {
        "messages": [
            {
                "metadata": {
                    "delivery": {
                        "queuedAt": "2021-03-16T18:07:16.000Z",
                        "totalSegments": 3
                    },
                    "requestId": "c223ecfe-46ef-4555-ac05-7b74b595f17d"
                },
                "attachments": [],
                "_id": "6050f3d2b507033e05e7ef06",
                "text": "Jackie here! Spring break might not be the same this year, but you can still enjoy a vacation at home! Plan your perfect staycation: https://www.dosomething.org/us/articles/how-to-plan-a-covid-19-staycation?utm_source=content_campaign&utm_medium=sms&utm_campaign=sms_pending_2021_03_16&user_id=5547be89469c64ec7d8b518d&broadcast_id=2IkRmKYUqySjPTEEHDS8q1",
                "direction": "outbound-api-send",
                "template": "autoReplyBroadcast",
                "conversationId": "6050ec27b507033e05e7ef01",
                "topic": "61RPZx8atiGyeoeaqsckOE",
                "userId": "5547be89469c64ec7d8b518d",
                "broadcastId": "2IkRmKYUqySjPTEEHDS8q1",
                "createdAt": "2021-03-16T18:07:14.833Z",
                "updatedAt": "2021-03-16T18:07:15.460Z",
                "__v": 0,
                "platformMessageId": "SM191fe426998e4689a4811419a4803b6a"
[...]

PR #554, opened 2021-03-16
Adds feature flag to toggle broadcastLite webhook configuration

aaronschachter:
### What's this PR do?

This pull request continues on #552, adding a `DS_GAMBIT_BROADCAST_NORTHSTARLESS` environment variable to toggle whether the API response of a `GET /v2/broadcasts/:id` should send broadcast messages to the new `broadcastLite` origin per https://github.com/DoSomething/blink/pull/257.

The response of the `GET /v2/broadcasts/:id` endpoint is used in Gambit Admin, allowing admins to copy/paste into Customer.io when sending a broadcast. When we're ready to go live with the broadcastLite route, we'll set the `DS_GAMBIT_BROADCAST_NORTHSTARLESS` config variable and flag for our admins that they'll need to copy/paste the new payload from Gambit Admin (not copy from an existing Customer.io webhook)

👀 

Last up is writing some integration tests for the `POST /v2/messages?origin=broadcastLite` route to verify the user is set as expected... but saving for another PR to keep these easier to review. By the time we get around to writing the integration test, it could just make sense to deprecate the old `broadcast` route that queries Northstar entirely...  

### Relevant tickets

#### Checklist
- [ ] Tested on staging.

### DIFF
--- a/lib/helpers/broadcast.js
+++ b/lib/helpers/broadcast.js
@@ -96,15 +96,22 @@ function isLegacyBroadcast(broadcast) {
  * getWebhookBodyForBroadcastId
  *
  * @param {string} broadcastId
+ * @param {boolean} isNorthstarless
  * @return {object}
  *
  * Returns the body of the POST request to Conversations /v2/messages?origin=broadcast.
  */
-function getWebhookBodyForBroadcastId(broadcastId) {
-  return {
-    northstarId: config.customerIo.userIdField,
+function getWebhookBodyForBroadcastId(broadcastId, isNorthstarless) {
+  const result = {
     broadcastId,
+    userId: config.customerIo.userIdField,
   };
+
+  return isNorthstarless ? Object.assign(result, {
+    addrState: config.customerIo.addrStateField,
+    mobile: config.customerIo.mobileField,
+    smsStatus: config.customerIo.smsStatusField,
+  }) : result;
 }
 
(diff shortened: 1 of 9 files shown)


lib/helpers/broadcast.js
```diff
@@ -96,15 +96,22 @@ function isLegacyBroadcast(broadcast) {
...
  *
  * @param {string} broadcastId
+ * @param {boolean} isNorthstarless
  * @return {object}
  *
  * Returns the body of the POST request to Conversations /v2/messages?origin=broadcast.
  */
-function getWebhookBodyForBroadcastId(broadcastId) {
-  return {
-    northstarId: config.customerIo.userIdField,
```
mendelB, line 105: Do we currently use this field from customer.io? Or is the `northstarId` -> `userId` change just future facing?

mendelB (approved):
👍

lib/helpers/broadcast.js
aaronschachter: Great Q! This `northstarId` was sent within the blink payload. https://github.com/DoSomething/blink/pull/257 modified Blink so that `northstarId` field is no longer required, and that we'll also accept a `userId` field. Once this is merged and deployed, I plan to make the `userId` field required on the Blink side (and remove support for the `northstarId` field)

PR #555, opened 2021-03-24
Use Northstar v2 API

aaronschachter:
### What's this PR do?

This pull request starts on changing all API requests to get, create, and update users to use Northstar's v2 endpoints instead of the v1 endpoints. When I test via the consolebot, things work as expected -- but it looks like we'll need to refactor the integration tests to mock authenticating with an oAuth token, as they're failing with `Cannot read property 'access_token' of undefined` errors:
https://github.com/DoSomething/gambit/blob/700b46e3fcbe98327384532bea8809743c4b7d7e/config/lib/northstar.js#L7-L11 

This isn't ready for review yet, but I'm opening for visibility / discussion on how best to move forward on this update. cc @mendelB @DFurnes 

* Eventually the `lib/northstar` and `lib/gateway` modules should be consolidated -- went this route for less tracked changes.

* Whenever this is ready for merge and then deployed, we'll need to update the `DS_NORTHSTAR_API_BASEURI` to remove the `v1` prefix. Because this config variable name is hardcoded in gateway-js, it seems easiest to just quickly make the variable change after deploy -- any errors we'd get from requests currently happening during the gap between deploy and variable change would be retried via Blink.

#### Checklist

- [ ] Tested on staging.


lib/northstar.js
```diff
@@ -1,96 +1,58 @@
...
 };
 
 /**
  * @param {string} id The Northstar id for this user
  * @return {Promise}
  */
-module.exports.unauthenticatedFetchUserById = function (id) {
-  logger.debug('fetchUserByIdUnauthenticated', { id });
-  return exports.getClient().Users.anonGet(config.getUserFields.id, id);
-};
```
aaronschachter, line 65: We no longer use this function -- at one point we were attempting to cache `GET /users/:id` responses in hopes of speeding up broadcasts but it didn't make much of a difference. Years later, we went with sending through the attributes from a Customer.io profile for Northstarless broadcasts in #552

lib/northstar.js
```diff
@@ -1,96 +1,58 @@
...
+module.exports.fetchUserById = async (id) => {
+  logger.debug('fetchUserById', { id });
 
-/**
- * @param {string} email
- * @return {Promise}
- */
-module.exports.fetchUserByEmail = function (email) {
```
aaronschachter, line 71: This is a relic from the early days of this project, where this app would handle messages from the Slack app. This feature was ported over to Gambit Slack, where we do fetch a user by email. It seemed at the time like a good time to separate non-critical functionality for staff use from the business-critical member-facing SMS app.

PR #569, opened 2021-06-21
Add additional logging for GraphQL fetches.

DFurnes:
### What's this PR do?

This pull request writes any (uncached) broadcasts or topics requests to the info log level. I'm hoping this will help us sort out why these endpoints are getting hit so much more heavily during broadcasts than we'd expect them to be.

I swapped the log level for this particular message in e877c8f.

I also set default `NODE_ENV` and `LOG_LEVEL` in b173608 to `.env.example`.

In particular, I'm curious to see how these requests are distributed between processes… maybe that'll give us a clue to why we're seeing so many of these requests go uncached. In theory, we should be fetching topics/broadcasts very infrequently so I think this is reasonable to log in production by default.

### Relevant tickets

#### Checklist
- [ ] Tested on staging.


weerd (approved):
👍

